PR #1, opened 2025-10-23
Small cleanups, mostly typing

fw-immunant:
Just from going through the repo and familiarizing myself. Should be no functional changes.

### DIFF
--- a/crisp/sandbox/docker.py
+++ b/crisp/sandbox/docker.py
@@ -39,6 +39,7 @@ class WorkContainer:
             #self.container.remove(v=True)
 
     def _checkout_tar_file(self, tar_bytes):
+        assert self.container is not None
         self.container.put_archive('/root/work/', tar_bytes)
 
     def checkout(self, n_tree):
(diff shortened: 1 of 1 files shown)


crisp/__main__.py
```diff
@@ -494,34 +496,26 @@ def main():
...
+        do_commit,
+        do_checkout,
+        do_cc_cmake,
+        do_transpile,
+        do_llm,
+        do_llm_repair,
+        do_test,
+        do_find_unsafe,
+        do_git,
+    ]}
```
kkysen: Python also has `match` now, but this is good, too.

fw-immunant:
> What command are you using to run typechecking on this? So far I've just been adding type annotations as documentation and haven't tried actually checking them. It would be nice to have a script or at least a line in the readme saying how to run it (and later we can add it to CI as well).

I've just been seeing errors in the Zed editor from the "basedpyright" LSP.

thedataking (approved):

PR #6, opened 2025-11-16
ci: run the `docker build`s from the `README.md`

kkysen:
Nothing is cached yet, but it'll be helpful to at least have something in CI to start with. The Docker containers aren't run at all yet either.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  docker-build-crisp:
+    runs-on: ubuntu-latest
```
spernsteiner: Would it make sense to pin this to a specific Ubuntu version for stability?

PR #7, opened 2025-11-16
ci: add job for building Rust tools

kkysen:


.github/workflows/ci.yml
```diff
@@ -22,3 +22,43 @@ jobs:
...
         run: docker build . -f Dockerfile.work -t tractor-crisp-user
+
+  build-tools:
+    runs-on: ubuntu-latest
+    
+    strategy:
+      matrix:
+        crate:
+          - find_unsafe
+          - split_ffi_entry_points
```
spernsteiner: I can definitely see myself forgetting to update this list when adding a new tool.  Maybe it would be better to have a single job that builds all the crates under `tools/` (we could set up a Cargo workspace to share syn/rust-analyzer builds if it becomes a performance issue), or else have a check that makes sure the list here includes everything in `tools/`.

spernsteiner (commented):
Please remove the rustfmt and clippy parts of this.  IME both tools are often more trouble than they're worth (e.g. I assume mixing together the joint and non-joint cases in `OutputBuffer::emit` was a clippy suggestion).  Otherwise this looks good.

.github/workflows/ci.yml
kkysen: I'm going to add this in a follow-up PR.

PR #10, opened 2025-11-17
ci: add `ruff format` and `ruff check`

kkysen:
This adds `ruff format --check` (formatting) and `ruff check` (linting) to CI, and also runs them on the codebase first to fix errors.  Almost all of these are automatically fixed, except for a couple of unused variables in ad68c42 that I fixed manually by prefixing them with `_`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,3 +33,5 @@ jobs:
       - run: uv sync
 
       - run: uv run ruff format --check
+
+      - run: uv run ruff check


crisp/llm.py
```diff
@@ -300,64 +314,69 @@ def get_default_model() -> str:
...
+        "messages": req_messages,
+        "model": model,
+    }
     resp = do_request(req, stream=True)
 
-    output = resp['choices'][0]['message']['content']
+    output = resp["choices"][0]["message"]["content"]
     output_files = input_code.files.copy()
-    files_changed = 0
+    _files_changed = 0
```
spernsteiner:
```suggestion
```

crisp/sandbox/sudo.py
```diff
@@ -41,53 +40,57 @@ def _popen_sudo(self, cmd, **kwargs):
 
     def start(self):
         # This command will error if the directory already exists.
-        p = self._run_sudo(('mkdir', self.dir_path))
+        _p = self._run_sudo(("mkdir", self.dir_path))
```
spernsteiner:
```suggestion
        self._run_sudo(("mkdir", self.dir_path))
```

kkysen (commented):
> > Why are you opposed to standardized formatting?
> 
> Following the major points of a common style guide is good, but trying to enforce every last minor point (as rustfmt does, and apparently ruff too) IME generally makes code less readable, not more. Whitespace is an important tool for conveying emphasis and structure. Automated tools have no idea what parts of the code are important and often expand and compress things in ways that make it harder to pick up on the overall shape of the code.
> 
> > makes it much easier to work as a team without conflicts about little formatting differences
> 
> I've never seen any notable disagreement over formatting on my other main project, which involves a large Haskell codebase with dozens of contributors over many years (and no official style guide, much less an autoformatter). The code isn't perfectly consistent, but everyone follows some kind of reasonable style (even if not everyone follows precisely the same one), so the code is all plenty readable.

You're able to turn off automated formatted in the few places where it's important to format it differently, but there's a reason why most collaborative projects use automated formatting.  I'm already finding it much more difficult to make changes without being able to format anything lest I create huge diffs, and we're bound to get significant git conflicts as different people format things differently, and this has repeatedly and consistently happened in every project I've worked on that didn't auto-format their code.

The Python code currently here is very un-idiomatic in a lot of places, and that makes it significantly harder to read at times.  Using standard formatting that programmers are used to reading would help a lot with this.  And then when I make changes, I now how to manually format them instead of auto-formatting it, and that means I have to spend less time on making meaningful semantic changes.  So I'll ask to please use standard formatting so that multiple people can more seamlessly contribute.

PR #14, opened 2025-11-24
docker: fix `crisp` installation

kkysen:
Previously, `uv sync` was installing `crisp` as an executable in certain local configurations, but not in Docker.  This installs it explicitly with `uv tool install .`, which I checked works in Docker, too. 

See https://github.com/GaloisInc/Tractor-Crisp/pull/13#discussion_r2551473585.


spernsteiner:
I cherry-picked this into #13, which is now merged.

PR #15, opened 2025-11-24
docker: remove unused toolchain components for transpiled code

kkysen:
Only `rustfmt` is needed.  See https://github.com/immunant/c2rust/blob/800158256678372b8abe15e471e78fc62d6fefdd/c2rust-transpile/src/build_files/generated-rust-toolchain.toml#L3.


spernsteiner (approved):
CI passes, so I guess these are indeed not required for building tools inside the container

PR #20, opened 2025-11-25
do_main: enable split_ffi

spernsteiner:
The split_ffi pass was implemented in #5, but it wasn't enabled at the time because it caused our success rate on weaker LLMs to drop from 50% to 10%.  However, it turns out that the original 50% number was an illusion: any library example containing only one function was being marked as successful immediately after running c2rust-transpile, without applying any safety refactoring.  This happened because all of the unsafe code was in one function and that function was marked `#[no_mangle]`, causing `find_unsafe` to ignore it.  With those test cases excluded, the original success rate was also around 10%, meaning split_ffi didn't actually impair performance.

This branch enables split_ffi so that safety refactoring will actually run on all examples, including single-function libraries.

### DIFF
--- a/tools/find_unsafe/src/main.rs
+++ b/tools/find_unsafe/src/main.rs
@@ -33,7 +33,7 @@ fn is_link_attr_meta(meta: &Meta) -> bool {
 fn is_link_attr_path(path: &Path) -> bool {
     match path.get_ident() {
         Some(i) => {
-            i == "no_mangle" || i == "link_name"
+            i == "no_mangle" || i == "export_name"
         },
         None => false,
     }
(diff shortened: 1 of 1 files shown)


crisp/__main__.py
```diff
@@ -296,6 +296,12 @@ def do_main(args, cfg):
         return
     w.accept(n_code, ('main', 'transpile'))
 
+    n_code = w.split_ffi(n_code)
+    if not w.test(n_code, n_c_code):
+        print('error: tests failed after split_ffi')
```
kkysen: nit: Printing errors to stderr can be more useful.

kkysen (approved):
LGTM from what I can tell.

crisp/__main__.py
spernsteiner: Leaving this as is for now.  We have a lot of prints in different places and I think it's best to have it all going to the same place, so that if stdout and stderr are redirected to different files, there's less risk of confusion about the ordering of messages.

spernsteiner:
Enabling split_ffi revealed a bug in find_unsafe: it was previously looking for either `#[no_mangle]` or `#[link_name]` to indicate that a function was an FFI entry point, but `#[link_name]` is only for imports, not exports.  It now looks for `#[export_name]` instead, which is the attribute applied by split_ffi.

PR #24, opened 2025-11-26
Run C2Rust on C datasets to get Rust datasets

souryadey:
- Introduce a script that can run C2Rust in bulk on datasets of C projects and produce datasets of Rust projects.
- Run this on 3 datasets and upload results:
  - `Public-Tests/B01_synthetic/` in the [test corpus](https://github.com/DARPA-TRACTOR-Program/Test-Corpus).
  - `Public-Tests/B01_organic/` in the [test corpus](https://github.com/DARPA-TRACTOR-Program/Test-Corpus).
  - `datasets/CBench` in [CRUST-Bench](https://github.com/anirudhkhatry/CRUST-bench), i.e. all the C projects that CRUST-Bench contains.

Rust datasets are uploaded as zip files. Unzipping will lead to one folder for each Rust project inside the dataset, and a `results.csv` file with details of which ones succeeded / failed.

All the test corpus C projects use CMake and all seem to have succeeded. On the other hand, the CRUST-Bench C projects use Make and the Makefile isn't always clean. About half of these failed, but the other half produced transpiled Rust files.

### DIFF
--- a/scripts/run_c2rust_and_get_datasets.py
+++ b/scripts/run_c2rust_and_get_datasets.py
@@ -19,6 +19,9 @@ import tempfile
 from tqdm import tqdm
 
 
+CONVERTED_RUST_PROJECTS_FOLDER = Path(os.path.dirname(os.path.realpath(__file__))).resolve().parent / 'converted_rust_projects'
+
+
 def run_subprocess_nodisp_check(*args, **kwargs) -> subprocess.CompletedProcess:
     """
     Run subprocess.run() with the defaults `check = True, stdout = subprocess.DEVNULL, stderr = subprocess.DEVNULL`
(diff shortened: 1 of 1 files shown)


converted_rust_projects/c2rust_CRUST-bench.zip
kkysen: How come these zip files need to be checked into git?

souryadey: Each zip file contains hundreds of files. For example, `c2rust_CRUST-Bench.zip` contains 50 folders inside it (one for each CRUST-Bench C project that could be transpiled using C2Rust), and each folder has a few Rust files and Cargo.toml inside it.

Happy to unzip and push them all if it sounds like a good idea.

PR #29, opened 2025-12-01
crisp: change `cc_cmake_op` to use `bear` to generate `compile_commands.json` instead of `cmake`

kkysen:
`cmake` can generate a `compile_commands.json` without building anything, but it includes everything, which may include duplicates.  If we build one target with `bear`, then we only get the commands used for that.

For now, this is folded into the existing `cc_cmake_op` instead of adding a separate `bear_op`.

### DIFF
--- a/crisp/analysis.py
+++ b/crisp/analysis.py
@@ -228,12 +228,13 @@ def _cc_cmake_impl(
     sb.checkout(c_code)
 
     exit_code = 0
-    logs = b''
+    logs = []
     for cmd in cmds:
         if exit_code != 0:
             break
         exit_code, new_logs = sb.run(cmd)
-        logs = b'\n\n'.join((logs, new_logs))
+        logs.append(new_logs)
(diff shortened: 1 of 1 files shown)


crisp/analysis.py
```diff
@@ -219,14 +219,26 @@ def inline_errors(
...
+    cmd = [*setup_cmd, "&&", *build_cmd]
+
     sb.checkout(c_code)
-    exit_code, logs = sb.run(cmd)
 
+    exit_code, logs = sb.run(setup_cmd)
+    if exit_code == 0:
+        new_exit_code, new_logs = sb.run(build_cmd)
+        exit_code = new_exit_code
+        logs += new_logs
```
spernsteiner: In other places I've been adding a `b'\n\n'` separator between old and new logs so they don't run together.

Since this is becoming a common pattern, maybe we should add a helper on `Sandbox` to run several commands in sequence.

PR #34, opened 2026-01-06
main: handle ref expressions like `a1b2c3.rust_code`

spernsteiner:
Currently, subcommands that expect a ref will accept either a tag name or a prefix (written in hexadecimal) of an existing `NodeId` hash.  This branch adds a third option: a ref followed by one or more attribute and/or indexing operations.

This is convenient when exploring the results of a CRISP run.  For example, suppose the logs mention `transpile_cc_op result = d7941...` and you want to see the code produced by that transpile operation.  Currently, you must first run `crisp show d7941`, which shows the metadata entry `'rust_code': NodeId(42f47...)`, and then separately run `crisp show --files 42f47`.  With this patch, you can run a single command `crisp show --files d7941.rust_code` to get the same effect.

### DIFF
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -21,8 +21,27 @@ from .work_dir import lock_work_dir, set_keep_work_dir
 from .workflow import Workflow
 
 
+ARG_PARSE_EPILOG = '''
+In subcommand arguments, a `NODE` can be:
+* A tag name, which refers to the most recent reflog entry for that tag
+* A hexadecimal ID, or any unique prefix of one
+* An expression `EXPR` that resolves to a node ID
+
+An `EXPR` can be:
+* Any `NODE`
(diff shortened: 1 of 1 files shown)


crisp/__main__.py
```diff
@@ -99,19 +101,68 @@ def parse_node_id_arg(mvir, s):
...
+                x = go(a.value)
+                idx = ast.literal_eval(a.slice)
+                return x[idx]
+            case ast.Attribute:
+                x = go(a.value)
+                node = mvir.node(x)
+                return getattr(node, a.attr)
+            case _:
+                print(a)
+                raise TypeError(f'unsupported expression kind: {a}')
```
kkysen: Do we both need to print it and include it in the error?  The latter seems sufficient to me.

spernsteiner: Removed the debug print in 9e7553ecf9d9fe2c858684ba7da225ad5335f799

PR #57, opened 2026-02-06
docker: update `c2rust` to `v0.22.0`

### LINKED ISSUE #46: Avoid /opt path dependency for c2rust-bitfields
If the final output of CRISP is copied out of the docker container, it won't build properly because it can't find `/opt/c2rust/c2rust-bitfields`.  Some possible fixes are:

* Always use `c2rust-bitfields` from crates.io.  (Currently we don't do this because our version of c2rust-transpile requires a `c2rust-bitfields` version that isn't yet published on crates.io.)
* Use a git dependency instead of a path dependency.  The downside of this is that Cargo will clone the entire C2Rust repo and all of its submodules, which is a bit slow.  But maybe we could pre-fetch this while building the docker image.
* Vendor `c2rust-bitfields` in a subdirectory of the output.

kkysen:
* Fixes #46.

This fixes the mismatch between `c2rust-bitfields` and the published `v0.21.0` version, so we don't need to use `--c2rust-dir` anymore.  `--c2rust-dir` still keeps things in sync going forward, but it removes portability out of the container.  Thus, this fixes #46.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ RUN uv python install
 RUN cd /opt \
     && git clone --depth 1 https://github.com/immunant/c2rust \
     && cd c2rust \
-    && git fetch --depth 1 origin e8d55cdc311912889ea82db6979c3709c7c8c4b2 \
+    && git fetch --depth 1 origin 1b2818697e2322b99ca0b78d8e23e8c965451343 \
     && git checkout FETCH_HEAD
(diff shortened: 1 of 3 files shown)


kkysen (commented):
@spernsteiner, could you test this however it's meant to be tested by T&E?

spernsteiner (approved):

PR #67, opened 2026-02-20
import `tools/` changes from `c2rust`

spernsteiner:
This copies the history of the `tools/` subdirectory as of https://github.com/immunant/c2rust/commit/05ffa6e39c86b1c3314d4afff2cd915d4e0b4b4e into this repo.  The goal of this move is to make it easier to develop the tools in tandem with the CRISP functionality that uses them.  The intent is that we will mainly develop these tools in this repo, and we'll copy updates back into the c2rust repo only when necessary (this is expected to be fairly rare).

This branch also adds a script to help with moving the history, in case it's useful in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,7 +123,10 @@ jobs:
       matrix:
         crate:
           - find_unsafe
+          - merge_rust
+          - related_decls
           - split_ffi_entry_points
+          - split_rust
 
     env:
       RUSTFLAGS: "-D warnings"


kkysen (commented):
We have `c2rust` as a submodule already.  Can we just use the tools in there directly?

spernsteiner:
That means every CRISP PR that touches the tools turns into two PRs, one in CRISP and another in c2rust.  Moving the tools into the CRISP repo makes it easier to make tool changes in connection with CRISP development, at the expense of making it harder to do changes in connection with c2rust.  Currently we expect to make a lot of changes for CRISP, as we experiment with pulling out various bits of information to feed to the LLM, but we expect to make almost no changes for C2Rust, since AIUI the current `split_rust`/`merge_rust` tools are sufficient for the needs of comment insertion.  So overall it's better to have them with CRISP, at least based on our current plans.

Dockerfile
```diff
@@ -88,6 +88,7 @@ RUN cd /opt/hayroll/hayroll \
 RUN ln -s /opt/hayroll/hayroll/hayroll /usr/local/bin/hayroll
 
 # Install CRISP tool binaries
+COPY tools/rust_util/ /opt/crisp-tools/rust_util/
```
kkysen: Does `split_ffi_entry_points` have a relative dependency on `rust_util`?

spernsteiner:
> Also, `c2rust` runs `cargo fmt --check` on the `tools/`, so I'd prefer to add that here, too, so that they don't get out of sync due to that.

Should be easy enough to `cargo fmt` as part of the transfer.

PR #80, opened 2026-03-02
workflow: use `hayroll`'s new `--keep-src-loc`s to run `c2rust-refactor` after `hayroll`

kkysen:
We'd like to integrate `hayroll` and `c2rust-refactor`.  However, `c2rust-refactor`'s `reorganize_definitions` requires annotations (like `#[c2rust::src_loc]` and `#[c2rust::header_src]`) from `c2rust-transpile`, which `hayroll` was removing as it wrapped `c2rust-transpile`.  How `hayroll` has a `--keep-src-loc` flag that preserves these annotations.  However, since `hayroll` handles conditional compilation, there may be different annotations on the same item that it has to merge, which it currently doesn't handle properly.  Thus, like for our other `hayroll`/`c2rust-transpile`/`c2rust-refactor` combinations, we first try `hayroll --keep-src-loc` + `c2rust-refactor` and fallback to the others if it fails.  This should hopefully cover most cases.

For a more complete solution, we may have to have `hayroll` wrap `c2rust-transpile` + `c2rust-refactor`, as `c2rust-refactor` can't maintain the conditional compilation that `hayroll` creates, as it's based on `rustc`, and `rustc` simply won't see those other `#[cfg]`s.  But we can start with this simpler solution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,9 @@ jobs:
       - name: Test on Test-Corpus (binary, runs c2rust transpile)
         run: ./scripts/test_eval.py Test-Corpus/Public-Tests/B01_synthetic/001_helloworld/
 
+      - name: Test on Test-Corpus (multi-file library, runs hayroll and c2rust refactor)
+        run: ./scripts/test_eval.py Test-Corpus/Public-Tests/B02_synthetic/char_to_bool/
+
       - name: Test on Test-Corpus (multi-file binary, runs c2rust transpile and c2rust refactor)
         run: ./scripts/test_eval.py Test-Corpus/Public-Tests/B01_synthetic/001_helloworld_with_header/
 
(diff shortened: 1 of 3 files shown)


fw-immunant:
Rebased.

fw-immunant:
``workflow: allow running c2rust-refactor after hayroll now that `--keep-src-loc` is supported`` and 
``workflow: deduplicate shared logic between c2rust-transpile and `hayroll`­`` are the only nontrivial changes, and only the former should change behavior.

spernsteiner:
Rebased, removed style changes, and added a few comments to clarify the output.  If CI passes and the logs show that it's exercising the new code path, then I think this is good to merge.

spernsteiner:
@fw-immunant Have you got any examples where Hayroll succeeds on a multi-file library?  I've tried several examples now (B02_organic/cJSON_lib, B02_organic/file_queue_lib, B02_organic/underhanded_c_nuke_lib, B02_synthetic/matrix_mult_lib) and none of them have worked.  It typically fails with a panic inside Hayroll:
```
thread 'main' (712) panicked at src/hayroll_ds.rs:338:22:
called `Option::unwrap()` on a `None` value
```

PR #94, opened 2026-04-08
Add partial support for multi-target C projects

spernsteiner:
This branch adds support in CRISP for multiple build artifacts.  The `crisp.toml` config file can contain several `[[transpile.artifacts]]` sections, each of which has its own `configure_cmds` (run prior to the build) and `build_cmds` (run under `bear` to capture `compile_commands.json`).  Each artifact gets transpiled into a separate crate, and CRISP adjusts the `Cargo.toml` files to combine all of the crates together into a workspace, such that `cargo build --release` at the top of the workspace will build all of the artifacts and put them in `target/release/` (as expected by the test infrastructure).

This branch also updates `test_eval.py` to use the new multi-artifact config options if it detects multiple targets in the CMake build system.  However, many of our tools (notably reorganize-definitions) don't support running over multiple crates at once.  As a temporary hack to keep things working, if `test_eval.py` detects a project with one binary and several library targets, it generates a config that builds the binary (which should also pull in the .c files for the libraries that the binary depends on) and then repackages the binary as one or more libraries.  The result is a binary and some libraries that each work in isolation, but don't work together properly (e.g. a downstream project that tried to link in two such libraries would fail with symbol conflicts).  The repackaging step is controlled by the config setting `transpile.artifacts[i].lib_from_bin_artifact`.  Once our other tools can handle multiple crates, we can change the `test_eval.py` script to generate a proper multi-artifact config instead.


crisp/analysis.py
```diff
@@ -266,7 +267,30 @@ def cc_cmake(cfg: Config, mvir: MVIR, c_code: TreeNode) -> CompileCommandsOpNode
...
+) -> CompileCommandsOpNode:
+    with run_sandbox(cfg, mvir) as sb:
+        work_dir = sb.join(cfg.relative_path('.'))
+        cc_path = sb.join(COMPILE_COMMANDS_PATH)
+        cmds = []
+        for cmd in art.configure_cmds:
+            cmd = f'cd {shlex.quote(work_dir)} && {cmd}'
+            cmds.append(['sh', '-c', cmd])
+        for cmd in art.build_cmds:
+            cmd = f'cd {shlex.quote(work_dir)} && bear --output {shlex.quote(cc_path)} -- {cmd}'
```
ahomescu: bear has this bug where if `{cmd}` contains a `--` then bear's argument parser will strip that out. In #88 I worked around that by wrapping the inner command in another `sh -c`.

crisp/workflow.py
```diff
@@ -238,27 +256,59 @@ def transpile(
...
+            art_code = mvir.node(n_op_transpile.rust_code)
+
+            # Patch Cargo.toml before building and testing.  This makes sure we
+            # test the code that will actually be used, and also gives
+            # `patch_cargo_toml` a chance to fix the c2rust-bitflags dependency if
+            # needed.
+            art_code = self.patch_cargo_toml(art_code, name = art_name)
+
+            # Hack: add -lcrypto, which is required for one test case
+            art_code = self.patch_build_rs(art_code, libs = ['crypto'])
```
ahomescu: This is only needed for the corpus, and I had to patch it out for libexpat. Could we make it optional somehow?

crisp/workflow.py
```diff
@@ -403,15 +458,18 @@ def transpile_cc_op(
...
         new_code = self.mvir.node(n_op.new_code)
         return new_code
 
     @step
-    def patch_cargo_toml_op(self, code: TreeNode) -> EditOpNode:
+    def patch_cargo_toml_op(self, code: TreeNode, name: str | None = None) -> EditOpNode:
         cfg, mvir = self.cfg, self.mvir
 
+        if name is None:
+            name = cfg.project_name
```
ahomescu, line 481: Alternatively, if `name is None`, keep the old names? I had to comment out this function for libexpat because it was renaming `runtests`.

crisp/workflow.py
```diff
@@ -445,6 +504,78 @@ def patch_cargo_toml_op(self, code: TreeNode) -> EditOpNode:
...
+                    'default-members': members,
+                },
+            }
+
+
+        new_files = code.files.copy()
+        for path, t in all_t.items():
+            new_files[path] = FileNode.new(mvir, toml.dumps(t)).node_id()
+
+        # Add a rust-toolchain.toml to the top-level workspace if needed.
```
ahomescu, line 619: If the top-level file is only a workspace, is this needed?

crisp/workflow.py
```diff
@@ -381,6 +381,7 @@ def transpile_cc_op(
                 if art_cfg.bin_main is not None:
                     c2rust_cmd.extend((
                         '--binary', art_cfg.bin_main,
+                        '--thin-binaries',
```
ahomescu, line 392: Just a quick fyi: I resolved all comments on https://github.com/immunant/c2rust/pull/1715, hoping to land it as soon as it gets approved.

crisp/analysis.py
```diff
@@ -434,26 +434,23 @@ def _split_rust_impl(
...
             return path
+        return 'src/lib.rs'
 
-    if (t_lib := t.get('lib')) is not None:
-        if (path := cargo_target_root_file(t_lib)) is not None:
+    for t_bin in t.get('bin', ()):
+        if (path := t_bin.get('path')) is not None:
             return path
+        if (name := t_bin.get('name')) is not None:
+            return f'src/bin/{name}.rs'
```
ahomescu, line 469: rustc looks at `src/bin/{name}/main.rs` too (but not `src/bin/{name}/mod.rs` like I initially thought)

crisp/workflow.py
```diff
@@ -506,6 +513,60 @@ def patch_cargo_toml_op(self, code: TreeNode, name: str | None = None) -> EditOp
...
+        t_lib = t['lib']
+        del t['bin']
+
+        # Use the baseline `lib` section to create the `lib` section for the
+        # derived artifact.
+        base_cargo_dir_rel = os.path.relpath(base_cargo_dir, new_cargo_dir)
+        t['lib'] = {
+            'path': os.path.join(base_cargo_dir_rel, t_lib['path']),
+            'name': name,
+            'crate-type': ['cdylib'],
```
ahomescu, line 557: Do we never see `staticlib` here?

crisp/workflow.py
```diff
@@ -506,6 +513,60 @@ def patch_cargo_toml_op(self, code: TreeNode, name: str | None = None) -> EditOp
...
+        t = toml.loads(base_cargo_toml.body_str())
+
+        t['package']['name'] = name
+
+        # The base Cargo.toml will have a `lib` containing all the relevant
+        # code, and a `bin` that wraps it and calls `lib::main`.  We discard
+        # the `bin` section, and read out the `lib` so we can create a modified
+        # version of it.
+        t_lib = t['lib']
+        del t['bin']
```
ahomescu, line 549: Is there always a single binary, or does this correctly delete an entire list?

crisp/workflow.py
```diff
@@ -258,42 +258,49 @@ def transpile(
...
+                # `patch_cargo_toml` a chance to fix the c2rust-bitflags dependency if
+                # needed.
+                art_code = self.patch_cargo_toml(art_code, name = art_name)
 
-            # Patch Cargo.toml before building and testing.  This makes sure we
-            # test the code that will actually be used, and also gives
-            # `patch_cargo_toml` a chance to fix the c2rust-bitflags dependency if
-            # needed.
-            art_code = self.patch_cargo_toml(art_code, name = art_name)
+                # Hack: add -lcrypto, which is required for one test case
```
ahomescu: Same comment from another commit applies here (not sure where it disappeared): this is only needed for the corpus and not for libexpat, could it be optional?

scripts/test_eval.py
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/env -S uv run
```
ahomescu, line 1: Is this a rename with edits? It might be easier to review if the rename was one commit, and the edits a separate one.

ahomescu, line 1: My general objection to this script is: if it's required or at least recommended to run it instead of CRISP directly (it commits the C code), maybe it should be integrated into CRISP? It might be useful to have a `crisp init` command instead.

scripts/test_eval.py
```diff
@@ -173,56 +141,120 @@ def relpath(path: Path, start: Path) -> Path:
...
+    # library of the appropriate name.  As long as the binary depends on the
+    # library in question, the library's source code will have been included in
+    # the binary, and the library compiled from that binary will export all of
+    # the right symbols.  However, it will also include the symbols from all
+    # the other libraries that got pulled into the binary, which means trying
+    # to link several such libraries into the same executable will fail.
+    #
+    # c2rust-transpile does have proper support for translating all of the
+    # libraries and binaries of a C project into separate Rust crates; once our
+    # other tools (such as c2rust-refactor) support multi-crate projects, we
```
ahomescu, line 262: Technically c2rust-refactor would work fine in that case, if you run it manually for every crate. It wouldn't work across crates, but I think that would take a lot of work to fix.

scripts/test_eval.py
```diff
@@ -173,56 +141,120 @@ def relpath(path: Path, start: Path) -> Path:
...
+    # should switch to that mode and get rid of this hack.
+
+    def is_binary(t):
+        return t['type'] == 'EXECUTABLE'
+    def is_library(t):
+        return t['type'] in ('STATIC_LIBRARY', 'SHARED_LIBRARY')
+    num_binaries = sum(1 for t in targets if is_binary(t))
+    num_libraries = sum(1 for t in targets if is_library(t))
+
+    # Supported combinations are 1 library and 0 binaries, or 1 binary and any
```
ahomescu, line 272: What about 1 library and N binaries? (especially with `--thin-binaries`)

scripts/test_eval.py
```diff
@@ -0,0 +1,258 @@
...
+    ]
+    commit_dirs = [
+        args.project_dir / "runner",
+        args.project_dir / "test_case",
+        args.project_dir / "test_vectors",
+        base_dir / "deployment",
+        base_dir / "tools",
+    ]
+
+    src_files = []
```
ahomescu: Since `crisp commit` only accepts a list of files and not directories, this part might be useful if integrated into CRISP itself so it works with other projects (like expat).

scripts/test_eval.py
```diff
@@ -325,7 +345,7 @@ def is_library(t):
                 rel_path = relpath(path, args.project_dir)
                 src_files.append(rel_path)
             for i in reversed(range(len(dirs))):
-                if dirs[i] in ("target", "__pycache__"):
+                if dirs[i] in exclude_dirs:
```
ahomescu: Is `crisp.toml` a file?

scripts/test_eval.py
```diff
@@ -341,11 +349,13 @@ def is_library(t):
...
         for root, dirs, files in start_dir.walk():
             for f in files:
+                if f in exclude_files:
+                    continue
                 path = root / f
                 rel_path = relpath(path, args.project_dir)
                 src_files.append(rel_path)
             for i in reversed(range(len(dirs))):
-                if dirs[i] in exclude_dirs:
+                if dirs[i] in exclude_subdirs or dirs[i].startswith('crisp-storage'):
```
ahomescu: I think a `pathspec` would be a lot more flexible here and avoid the need for this kind of hack. See my agent PR for an example, you can do something like `**/crisp-storage*/**` in gitignore format.

scripts/test_eval.py
```diff
@@ -276,19 +276,42 @@ def is_library(t):
...
+            # `/tmp/sandbox/test_case` or similar, and attempting to build
+            # `foo` fails because the target is now called `sandbox`.  It
+            # happens to be the case that all such projects in `Test-Corpus`
+            # have only a single target.  We handle these projects by building
+            # without a specific target name, which builds the sole target
+            # regardless of its name, while still setting the artifact name in
+            # `crisp.toml` to the original name detected by this script, so
+            # that the final library or binary will still have the name
+            # expected by the test suite.
+            build_cmd.extend(('--', name))
```
ahomescu, line 297: See my other comment: bear will eat up this `--`.

crisp/workflow.py
spernsteiner: I thought the toolchain config had to be in the directory from which you run `cargo` (or a parent directory), since the `rustup` wrapper has to decide which toolchain to use before any Cargo logic can run.  So I would expect that `cd workspace && cargo build` would respect `workspace/rust-toolchain.toml` but not any `workspace/pkg1/rust-toolchain.toml` files.

crisp/workflow.py
spernsteiner: Yes, I'm waiting on that so I can bump the c2rust submodule to point to the final merge commit

crisp/workflow.py
spernsteiner: `t` will look like
```Python
t = {
    'bin': [
        bin1,
        bin2,
        bin3,
    ],
    'other-keys': 'other-values',
}
```
So `del t['bin']` will delete all binaries.

crisp/analysis.py
spernsteiner: IMO this code is already getting annoyingly complicated.  I'll add a comment saying that if we need to handle any more cases beyond this, we should switch to calling `cargo metadata` or something, and let Cargo do all the name/path resolution properly.

crisp/workflow.py
spernsteiner: This is generating the `lib` entry from scratch, so we can set `crate-type` to whatever we want.  So far we've only needed `cdylib`s.  (This is somewhat an artifact of T&E's current testing setup, which wants to `dlopen` the library it's testing.)

scripts/test_eval.py
spernsteiner: The rename is a separate commit from the edits, but the changes are so extensive that looking at just the edits didn't actually help that much IMO.  But if you want, you can click the "All commits" dropdown at the top of github's diff viewer and restrict it to show only the post-rename parts.

spernsteiner: Currently all `crisp` subcommands expect you to already have a `crisp.toml` in place, but the purpose of this script is to generate the `crisp.toml`.

Also, this particular script is not mandatory for running CRISP.  It's specific to CMake and to Test-Corpus in particular.  For other projects, you'd likely want to write the config by hand or use a different script (e.g. at some point we might want to write one that generates configs for autotools projects).

scripts/test_eval.py
spernsteiner: The specific issue is that reorganize-definitions won't resolve `extern "C" { fn foo(); }` to `use other_crate::foo;`.  This also means it won't unify any structs or other types that `foo` might mention in its signature, so you'll end up with a separate copy of the struct in each crate.

scripts/test_eval.py
spernsteiner: AIUI that would require translating all binaries in a single c2rust-transpile call so that they all share the same library, which in turn requires using the cc/ld wrappers instead of `bear`.

scripts/test_eval.py
spernsteiner: Yes, but I don't understand the question.  Are you asking whether we should add `crisp.toml` to `exclude_files`?

crisp/workflow.py
spernsteiner: You can now set the exact name you want for each artifact in the `[[transpile.artifacts]]` section, and it will use that here.  IIRC the name in the Cargo.toml usually gets set initially to something unhelpful like `translated_rust` or `hayroll_out`, so we basically always need to rename it somehow.

crisp/analysis.py
spernsteiner: Wrapped the command in `sh -c` in 0d9e0cfee49d8e16e543c1db138410b4c0b40457

crisp/analysis.py
spernsteiner: Added a TODO comment in 58f1422ee3a46c99048f978df9ddee6cd414823b

crisp/workflow.py
spernsteiner: Added a config option in a58e9f4f096f1ee2f839966431f9cebd749d50a7 to control which libraries to add here (if any)

crisp/workflow.py
spernsteiner: Answered in this thread: https://github.com/GaloisInc/Tractor-Crisp/pull/94#discussion_r3054953889

scripts/test_eval.py
spernsteiner: Replied to the other comment https://github.com/GaloisInc/Tractor-Crisp/pull/94#discussion_r3054886210

crisp/workflow.py
ahomescu: I think I know what you mean, `--manifest-path` has that problem too.

crisp/workflow.py
ahomescu: It has landed now.

scripts/test_eval.py
ahomescu: It looks like maybe just a GitHub problem, I expected it to show me that commit as only a file rename without any changes.

scripts/test_eval.py
ahomescu: At first I was going to say that would be hard to do across crates, but rustc does have all exports and their types/signatures in its crate metadata for all imported crates (transitively). We probably could do this specific thing.

scripts/test_eval.py
ahomescu: That makes sense. The other thing I think might still be useful (see my other comment below) is either a script or `crisp` subcommand that commits the original C code (unless I missed something that's already there).

scripts/test_eval.py
ahomescu: Exactly, it seemed wrong to put it in `exclude_dirs`.

ahomescu (approved):
I think the only significant issue remaining is having `crisp.toml` in `exclude_subdirs`, once that's resolved we can merge this.

crisp/workflow.py
spernsteiner: Updated the submodule pointer in eb77f19e14d60bea59025005b9223fe9870bcac1

scripts/test_eval.py
spernsteiner: The metadata for dependencies includes info about their exports, but what about imports?  Example:

* `lib1` exports `fn f(ptr: *const lib1::MyStruct)`
* `lib2` imports `fn f(ptr: *const lib2::MyStruct)` and exports `fn g(ptr: *const lib2::MyStruct)`
* `lib3` imports `fn g(ptr: *const lib3::MyStruct)`

When running on `lib3`, there's enough information to recognize that `lib3::g == lib2::g` and thus `lib3::MyStruct == lib2::MyStruct`, but is there enough to know that `lib2::MyStruct == lib1::MyStruct`?

If not, we could maybe do a sort of multi-step solution that first collects a set of constraints from each crate, then solves the constraints, and finally rewrites each crate to apply them.  (We might need something like this anyway if the metadata isn't sufficient to let us find and rewrite the relevant source files for the dependencies.)

scripts/test_eval.py
spernsteiner: I'll try extending `crisp commit` to accept both files and directories and to have a `--exclude` option.  That will cover most of this logic and make it reusable for other projects.

scripts/test_eval.py
spernsteiner: I'm extending `crisp commit` to help somewhat with this: https://github.com/GaloisInc/Tractor-Crisp/pull/94#discussion_r3065795293

scripts/test_eval.py
spernsteiner: Extended `crisp commit` in 1736aab23cba5bae6f8c9a22cdd3a867dbc8e5fe

scripts/test_eval.py
spernsteiner: As of 1736aab23cba5bae6f8c9a22cdd3a867dbc8e5fe, we now use `crisp commit` with `--exclude crisp*` (among other `--exclude` options) to cover both `crisp.toml` and the `crisp-storage/` directory.

scripts/test_eval.py
spernsteiner: As of 1736aab23cba5bae6f8c9a22cdd3a867dbc8e5fe, this script uses `crisp commit --exclude PATTERN`, which is implemented in terms of `PathSpec`.

PR #95, opened 2026-04-10
workflow: pass --add-unsafe-blocks to split_ffi

spernsteiner:
In #93 we disabled adding `unsafe` blocks inside the `unsafe fn`s generated by `split_ffi`, since a future version of c2rust-transpile (after https://github.com/immunant/c2rust/pull/1720) won't require them.  But the current version does still require them, so we need to keep generating them for now.

### DIFF
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -600,7 +600,8 @@ class Workflow:
         with run_sandbox(cfg, mvir) as sb:
             sb.checkout(n_tree)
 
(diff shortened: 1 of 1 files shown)


spernsteiner:
Turns out this doesn't apply to current `main`, since we haven't yet upgraded to a version of C2Rust that sets `deny(unsafe_op_in_unsafe_fn)` (that will happen as part of #94)

PR #96, opened 2026-04-12
Add --codex-login flag for alternative agent authentication

thedataking:
- Add `--codex-login` CLI flag that uses the host's `codex login` session (`~/.codex/auth.json`) instead of `CRISP_API_KEY` for agent authentication. Requires `--llm-mode=agent`.

### DIFF
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -62,6 +62,9 @@ def parse_args():
         choices=('default', 'no_ffi', 'agent', 'agent_sim_no_tests'),
         default='default',
         help='which style of LLM-based rewriting to use')
+    main.add_argument('--codex-login', action='store_true',
+        help="use the host's `codex login` credentials instead of CRISP_API_KEY; "
+            "requires --llm-mode=agent or --llm-mode=agent_sim_no_tests")
 
     safety_loop = sub.add_parser('safety-loop')
     safety_loop.add_argument('--c-code', default='c_code')
@@ -70,6 +73,9 @@ def parse_args():
         choices=('default', 'no_ffi', 'agent', 'agent_sim_no_tests'),
         default='default',
         help='which style of LLM-based rewriting to use')
+    safety_loop.add_argument('--codex-login', action='store_true',
+        help="use the host's `codex login` credentials instead of CRISP_API_KEY; "
+        "requires --llm-mode=agent or --llm-mode=agent_sim_no_tests")
 
     repl = sub.add_parser('repl')
     repl.add_argument('--node', '-n', action='append', metavar='[NAME=]NODE',
@@ -119,7 +125,10 @@ def parse_args():
     sandbox_run.add_argument('--checkout', action='append', default=[], metavar='NODE',
         help='check out files from this node into the sandbox')
 
-    return ap.parse_args()
+    args = ap.parse_args()
+    if getattr(args, 'codex_login', False) and getattr(args, 'llm_mode', None) not in ('agent', 'agent_sim_no_tests'):
+        ap.error('--codex-login requires --llm-mode=agent or --llm-mode=agent_sim_no_tests')
+    return args
 
 
 def parse_node_id_arg(mvir, s):
@@ -207,7 +216,7 @@ def parse_node_id(mvir, s):
 
 def do_main(args, cfg):
     mvir = MVIR(cfg.mvir_storage_dir, '.')
-    w = Workflow(cfg, mvir)
+    w = Workflow(cfg, mvir, codex_login=args.codex_login)
 
     c_code_node_id = parse_node_id_arg(mvir, args.node)
     n_c_code = mvir.node(c_code_node_id)
@@ -359,7 +368,7 @@ def safety_loop_common(args, cfg, mvir, w, n_code, n_c_code):
 
 def do_safety_loop(args, cfg):
     mvir = MVIR(cfg.mvir_storage_dir, '.')
-    w = Workflow(cfg, mvir)
+    w = Workflow(cfg, mvir, codex_login=args.codex_login)
 
     c_code_node_id = parse_node_id_arg(mvir, args.c_code)
     n_c_code = mvir.node(c_code_node_id)
--- a/crisp/agent.py
+++ b/crisp/agent.py
@@ -2,8 +2,9 @@
 Rewrite operations using AI agent tools, such as codex-cli
 """
 
-from pathspec.pathspec import PathSpec
 import os
+
+from pathspec.pathspec import PathSpec
 
 from . import llm
 from .config import Config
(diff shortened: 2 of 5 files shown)


crisp/sandbox/docker.py
```diff
@@ -40,7 +40,7 @@ def stop(self):
             self.container.stop(timeout=1)
             #self.container.remove(v=True)
 
-    def _checkout_tar_file(self, tar_bytes):
+    def checkout_tar_file(self, tar_bytes):
```
spernsteiner: Since this is currently being used only to check out a single file, I think it would be cleaner to have a function like `checkout_file_untracked` that works like `checkout_file` but takes the file body as `bytes` rather than as a `FileNode` (and then have `checkout_file` dispatch to `_untracked(n_file.body())`).  This would be easier to implement in `sandbox.bwrap`, which doesn't have an equivalent of `checkout_tar_file`.

crisp/__main__.py
```diff
@@ -197,6 +200,11 @@ def parse_node_id(mvir, s):
 
 
 def do_main(args, cfg):
+    if args.codex_login and args.llm_mode != 'agent':
+        print('--codex-login implies --llm-mode=agent; switching from %r to agent'
+              % args.llm_mode)
+        args.llm_mode = 'agent'
```
spernsteiner: This prevents using `--codex-login` with `--llm-mode=agent_no_tests`.  Maybe this code should check that `llm_mode` is one of the agent modes, but otherwise leave the setting unchanged?

crisp/__main__.py
```diff
@@ -234,11 +242,11 @@ def do_main(args, cfg):
...
-    n_code = w.split_ffi(n_code)
-    if not w.test(n_code, n_c_code):
-        print('error: tests failed after split_ffi')
-        return None
-    w.accept(n_code, ('main', 'split_ffi'))
+    #n_code = w.split_ffi(n_code)
+    #if not w.test(n_code, n_c_code):
+    #    print('error: tests failed after split_ffi')
+    #    return None
+    #w.accept(n_code, ('main', 'split_ffi'))
```
spernsteiner: Uncomment?  Or maybe add a `--no-split-ffi` option, if you need to skip this part for expat?

crisp/__main__.py
```diff
@@ -252,7 +260,8 @@ def safety_loop_common(args, cfg, mvir, w, n_code, n_c_code):
         try:
             match args.llm_mode:
                 case 'agent':
-                    n_new_code = w.agent_safety(n_code, n_c_code)
+                    n_new_code = w.agent_safety(n_code, n_c_code,
+                        codex_login=args.codex_login)
```
spernsteiner: The goal of `Workflow` is to minimize the number of configuration arguments needed to make a call, so I'd rather avoid the `codex_login` arg here if possible.  Some options:
* Pass `codex_login` to the `Workflow` constructor and save it internally alongside `cfg` and `mvir`
* Make it an env var, like the way we currently handle `$CRISP_API_KEY` and such
* Add it to the `Config` somewhere (not sure if this makes sense given none of the other API settings are handled that way)

crisp/workflow.py
```diff
@@ -255,10 +255,10 @@ def transpile(
         # test the code that will actually be used, and also gives
         # `patch_cargo_toml` a chance to fix the c2rust-bitflags dependency if
         # needed.
-        code = self.patch_cargo_toml(code)
+        ###code = self.patch_cargo_toml(code)
 
         # Hack: add -lcrypto, which is required for one test case
-        code = self.patch_build_rs(code, libs = ['crypto'])
+        ###code = self.patch_build_rs(code, libs = ['crypto'])
```
spernsteiner: Uncomment.  In #94 (which I'm hoping to merge today) we add a config option to control the `patch_build_rs` part, and we also make changes to `patch_cargo_toml`  that will hopefully work better for expat.

crisp/workflow.py
```diff
@@ -317,6 +317,7 @@ def transpile_cc_op(
                     "--output-dir",
                     sb.join(output_path),
                     "--emit-build-files",
+                    "--thin-binaries",
```
spernsteiner: Shouldn't be part of this PR (but #94 also enables this)

crisp/agent.py
```diff
@@ -52,14 +103,17 @@ def run_rewrite(
         sb.checkout(input_code)
         sb.checkout(test_code)
 
+        if codex_login:
+            _inject_codex_auth(sb)
```
spernsteiner, line 108: This should print a warning similar to the `CRISP_API_KEY` leakage warning above

crisp/__main__.py
ahomescu: I agree that it shouldn't change the setting; I would make it a hard error instead.

crisp/agent.py
```diff
@@ -13,25 +18,70 @@
...
+def _codex_command(subcmd: str, args: list[str], codex_login: bool = False) -> list[str]:
     cmd = ['codex', subcmd]
-    for k, v in config_settings.items():
-        cmd += ['-c', f'{k}={v}']
-    cmd += ['--profile', 'crisp']
-    cmd += args
 
+    if codex_login:
+        # Use the host's `codex login` credentials (auth.json).  We only
+        # override the model; everything else uses codex's defaults.
```
ahomescu, line 33: Why are we overriding the model though? Couldn't we use the one here together with the `auth.json` from the host?

crisp/agent.py
```diff
@@ -13,25 +18,70 @@
...
+def _inject_codex_auth(sb):
+    """Copy the host's ``~/.codex/auth.json`` into the container's work
+    directory so that codex-cli can authenticate using the host's
+    ``codex login`` session.
+
+    The file only lives for the lifetime of the container and is never
+    written to MVIR.  The ``.codex/`` ignore pattern in ``run_rewrite``
+    also ensures it is excluded from ``commit_dir`` output.
+    """
+    host_auth = os.path.expanduser('~/.codex/auth.json')
```
ahomescu: Check `CODEX_HOME` as well?

crisp/workflow.py
ahomescu: The base branch here is wrong, Per is going to rebase to `main`.

crisp/workflow.py
ahomescu: Same here: base branch issue.

crisp/__main__.py
```diff
@@ -66,6 +66,9 @@ def parse_args():
     safety_loop.add_argument('--llm-mode', choices=('default', 'no_ffi', 'agent', 'agent_no_tests'),
         default='default',
         help='which style of LLM-based rewriting to use')
+    main.add_argument('--codex-login', action='store_true',
```
ahomescu: We also need the argument for `safety_loop`, otherwise `main` will re-transpile.

crisp/__main__.py
```diff
@@ -197,6 +200,11 @@ def parse_node_id(mvir, s):
 
 
 def do_main(args, cfg):
+    if args.codex_login and args.llm_mode != 'agent':
```
ahomescu: Add it to `do_safety_loop`? Or better, move it to `safety_loop_common`?

crisp/__main__.py
ahomescu: Speaking of, I just noticed that even for `main` this is in the wrong place.

crisp/__main__.py
thedataking: Resolved via rebase on `main`.

crisp/sandbox/docker.py
thedataking: That's cleaner indeed. Done. 

(It didn't sound like you wanted me to go update the other sandboxes. Correct me if I'm wrong.)

crisp/agent.py
thedataking: `codex` does not support snapshot-style model names. I've added a comment explaining this.

crisp/__main__.py
thedataking: Good point. Hoisted it into `parse_args` instead, using `ap.error(...)` for argparse's standard usage+exit behavior. So neither `do_main`, `do_safety_loop`, or `safety_loop_common` has to care about this.

ahomescu (approved):

ahomescu:
LGTM and Per addressed all the previous comments, going to approve it so we can land it. If there is more feedback, we can address it in a follow-up.

crisp/sandbox/docker.py
spernsteiner: Sorry, I thought I already replied to this, but I'm not sure where that comment ended up.  In short: please make sure that if you run with `--codex-login` and an unsupported sandbox, it fails with an obvious error indicating that it's unsupported (or else make it work with all sandboxes).  If it fails with "method not found" (which I think it currently does) then that could look like bitrot or a bug instead of an expected limitation (in 6 months when we've all forgotten the details)

PR #97, opened 2026-04-13
ci: run tests using docker image from the current branch, not registry

spernsteiner:
CI on #94 was failing (e.g. [here](https://github.com/GaloisInc/Tractor-Crisp/actions/runs/24354701260/job/71118437606)) because it was using an outdated version of the tractor-crisp-user container to run tests.  This previously went undetected because any time we ran a tool inside the container, the arguments to the tool were compatible with both the new version and the old one.  But #94 started passing a new flag to `find-unsafe` that wasn't supported by the old version, causing CI to fail.

Codex (GPT-5.4) proposed using `docker buildx build --load` and removing the `docker pull` step so that the later steps use the image built by the earlier ones, without going through the container registry.  There was a comment saying that the current setup avoids using `--load` to keep the runner from running out of disk space, but the version with `--load` seems to work fine on #94.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,9 +48,6 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y \
             universal-ctags
-      
-      # `--push` instead of `--load` so that everything is cached first
-      # and so we don't accidentally run out of storage.
 
       - name: Set $IMAGE_PREFIX
         run: |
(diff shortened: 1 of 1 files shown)


kkysen (commented):
This is what I had initially tried (`--load`) and was often failing by running out of storage, and is also sensitive to changes in the Dockerfile that might use more storage, which is pretty easy to do.

I think it also messes with caching.  I forget exactly how, but it might mean either it's not cached at all anymore, or only cached at the end if there are no errors vs. caching each layer as each layer succeeds.

spernsteiner:
I'm going to merge this even though it might still be broken, because at least it seems less broken than the current setup, and it should unblock #94.

PR #98, opened 2026-04-14
deps/hayroll: update to include latest fixes for `lib.rs` pragma merging instead of hardcoding

kkysen:
This should fix some `Test-Corpus` test cases, such as when we produce `&raw`s (`#![feature(raw_ref_op)]` on the old nightly.  I observed this happening in `Test-Corpus/Public-Tests/B02_synthetic/arity_lib`, which now successfully works with `hayroll`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,19 +59,19 @@ RUN cd /opt/c2rust \
 # Note that Hayroll's `prerequisites.bash` pins its git dependencies to
 # specific tags, so we don't have to worry (much) about ensuring we get the
 # right version.
-COPY deps/hayroll /opt/hayroll/hayroll
+COPY deps/hayroll /opt/hayroll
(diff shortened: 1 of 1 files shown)


spernsteiner:
The docker changes look plausible, but I'd like to manually compare the results before and after to see what's changed with this newer version of Hayroll.  Do you know of any specific Test-Corpus examples that would demonstrate the differences?

kkysen (commented):
> The docker changes look plausible, but I'd like to manually compare the results before and after to see what's changed with this newer version of Hayroll. Do you know of any specific Test-Corpus examples that would demonstrate the differences?

I tested on `Test-Corpus/Public-Tests/B02_synthetic/arity_lib` with `hayroll`.  Previously, it would fail with `error[E0658]: raw address of syntax is experimental`, but after this update it works.

spernsteiner (approved):
I tested on arity_lib with and without this change and saw that it fixes the error as you described.

PR #99, opened 2026-04-14
Misc fixes (2026-04-13)

spernsteiner:
* Include `test_eval.py` in the docker container, so we can use it when running all of CRISP under docker.
* Use `find-unsafe` from `$PATH` instead of building it with Cargo on demand.  The build was failing because we don't copy the `tools/` directory into the full `tractor-crisp` container.  However, `find-unsafe` is already installed into `$PATH` within the sandbox, so it seemed easiest to switch to running `find-unsafe` in the sandbox the same way we run all of the other helper tools.
* Add a new `LLM_SAFETY_MAX_CONSECUTIVE_FAILURES` env var.  When using `--llm-mode agent` on large projects, it's useful to set `LLM_SAFETY_TRIES` very high (e.g. 100 or more), since it edits only a small portion of the codebase on each run.  But this could waste a lot of tokens if the agent gets stuck and repeatedly runs without making progress.  This new option cuts off the loop if the agent goes too many iterations without improving the unsafe count beyond its previous best.
* Update `Dockerfile` to install `ctags` (needed for running `test_eval.py`) into the `tractor-crisp` container.  This isn't needed in `tractor-crisp-user` since it's used for running CRISP itself, not for running any of the sandboxed tools that CRISP calls out to.
* Update `Dockerfile` to add the Cargo sparse registry config to `~crisp_sandbox_user/.cargo/config.toml`.  Currently we put this setting in `$CARGO_HOME/config.toml`, which for `root` expands to `/usr/local/cargo/config.toml`, but `crisp_sandbox_user` doesn't have `$CARGO_HOME` set in its environment and defaults to `~/.cargo/config.toml` instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -117,6 +117,11 @@ RUN useradd -m crisp_sandbox_user
 ENV CRISP_SANDBOX=sudo
 ENV CRISP_SANDBOX_SUDO_USER=crisp_sandbox_user
 
+# Enable sparse registry for the sandbox user
+RUN mkdir -v /home/crisp_sandbox_user/.cargo \
+    && cp -v $CARGO_HOME/config.toml /home/crisp_sandbox_user/.cargo/config.toml \
+    && chown -Rv crisp_sandbox_user:crisp_sandbox_user /home/crisp_sandbox_user/.cargo
+
 # CRISP setup.  This comes last because it changes the most often.
 WORKDIR /opt/tractor-crisp
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,6 +103,9 @@ RUN echo "export PATH=$PATH:/usr/local/cargo/bin" >>/etc/profile
 
 FROM tractor-crisp-user AS tractor-crisp
 
+# Additional dependencies for scripts
+RUN apt-get install -y universal-ctags
+
 # Set up sudo so CRISP can use it for sandboxing
 RUN apt-get install -y sudo
 RUN sed -i -e 's,secure_path=",&/usr/local/cargo/bin:,' /etc/sudoers
--- a/Dockerfile
+++ b/Dockerfile
@@ -119,6 +119,7 @@ WORKDIR /opt/tractor-crisp
 
 COPY pyproject.toml uv.lock ./
 COPY crisp/ ./crisp/
+COPY scripts/test_eval.py ./scripts/test_eval.py
 RUN uv sync
 # FIXME: currently disabled in favor of the wrapper script below.  Some parts
 # of CRISP use `os.path.dirname(__file__)` to find related files, but `uv`
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,7 +103,7 @@ RUN echo "export PATH=$PATH:/usr/local/cargo/bin" >>/etc/profile
 
 FROM tractor-crisp-user AS tractor-crisp
 
-# Additional dependencies for scripts
+# Dependencies for `scripts/test_eval.py`
 RUN apt-get install -y universal-ctags
 
 # Set up sudo so CRISP can use it for sandboxing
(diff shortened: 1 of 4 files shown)


crisp/analysis.py
```diff
@@ -371,43 +371,41 @@ def crisp_git_state(subdir=None) -> str:
...
-        print(' --- stderr ---\n%s\n' % p.stderr.decode('utf-8', errors='replace'))
-        p.check_returncode()
+    cmd_wrapped = ['sh', '-c', shlex.join(cmd) + ' >unsafe.json']
+    exit_code, logs = sb.run(cmd_wrapped)
 
-    # Check that stdout is valid json
-    _ = json.loads(p.stdout.decode('utf-8'))
+    if exit_code == 0:
+        if isinstance(logs, bytes):
+            logs = logs.decode('utf-8')
```
kkysen: nit:

```suggestion
            logs = logs.decode()
```

This is the default already.

crisp/__main__.py
```diff
@@ -240,12 +240,38 @@ def do_main(args, cfg):
     safety_loop_common(args, cfg, mvir, w, n_code, n_c_code)
 
 def safety_loop_common(args, cfg, mvir, w, n_code, n_c_code):
+    # Try at most this many times in total to make the code safe.
     llm_safety_tries = int(os.environ.get("LLM_SAFETY_TRIES", "3"))
```
ahomescu, line 244: If someone wanted to run this like I did for libexpat (until `unsafe_count == 0`), should they just put in a really large number here like 1000000? We could add a special `INFINITE` value instead.

spernsteiner:
Added points 4 and 5 to the PR description, covering changes that were needed to get CRISP and `test_eval.py` working inside the `tractor-crisp` container.

Dockerfile
```diff
@@ -103,6 +103,9 @@ RUN echo "export PATH=$PATH:/usr/local/cargo/bin" >>/etc/profile
 
 FROM tractor-crisp-user AS tractor-crisp
 
+# Additional dependencies for scripts
```
kkysen:
```suggestion
# Dependencies for `scripts/test_eval.py`
```

Dockerfile
```diff
@@ -114,11 +117,17 @@ RUN useradd -m crisp_sandbox_user
 ENV CRISP_SANDBOX=sudo
 ENV CRISP_SANDBOX_SUDO_USER=crisp_sandbox_user
 
+# Enable sparse registry for the sandbox user
+RUN mkdir -v /home/crisp_sandbox_user/.cargo \
```
ahomescu:
```suggestion
RUN mkdir -v /home/$CRISP_SANDBOX_SUDO_USER/.cargo \
```

in case we ever want to change the variable value? I might be getting the syntax slightly wrong, but I know there's a way to reference `ENV` variables from `RUN`.

Dockerfile
```diff
@@ -117,6 +117,11 @@ RUN useradd -m crisp_sandbox_user
...
+    && cp -v $CARGO_HOME/config.toml /home/crisp_sandbox_user/.cargo/config.toml \
+    && chown -Rv crisp_sandbox_user:crisp_sandbox_user /home/crisp_sandbox_user/.cargo
```
kkysen:
```suggestion
RUN mkdir -v /home/$CRISP_SANDBOX_SUDO_USER/.cargo \
    && cp -v $CARGO_HOME/config.toml /home/$CRISP_SANDBOX_SUDO_USER/.cargo/ \
    && chown -Rv $CRISP_SANDBOX_SUDO_USER:$CRISP_SANDBOX_SUDO_USER /home/$CRISP_SANDBOX_SUDO_USER/.cargo
```

Shouldn't we use `$CRISP_SANDBOX_SUDO_USER`?

kkysen (approved):
LGTM outside of a few minor things.

crisp/__main__.py
spernsteiner: Yeah, I think the easiest approach for now is to just you should just set it to a big number.  In general these env var options seem a little clunky, so we may want to rework them completely at some point.

kkysen (approved):

spernsteiner:
I'm waiting on a final test run on P01, and once it's done I plan to merge this

PR #104, opened 2026-04-17
Better handling of the case where no tests are available

spernsteiner:
This branch adds the ability to leave `test_command` unset in the config file, and modifies `test_eval.py` to leave it unset when run with `--no-runtests` or `--no-git`.  Leaving it unset causes CRISP to skip running tests, and changes the agent prompt to avoid mentioning them.  This lets us use `--llm-mode agent` (with `test_command` unset) for the evaluation, instead of `--llm-mode agent_no_tests` (which is more of a debugging feature, and has been renamed to `agent_sim_no_tests` to make that a little clearer).

The specific issue we were seeing is that with `test_eval.py --no-runtests -- --llm-mode agent_no_tests`, if the agent produces code that doesn't compile, the entire CRISP loop crashes and produces no output.  This is a combination of `--no-runtests` defaulting to `test_command = "cargo build"`, since previously it wasn't legal to omit it, and `--llm-mode agent_no_tests` asserting that tests must always succeed as a debugging mechanism.

### DIFF
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -58,14 +58,16 @@ def parse_args():
 
     main = sub.add_parser('main')
     main.add_argument('node', nargs='?', default='c_code')
-    main.add_argument('--llm-mode', choices=('default', 'no_ffi', 'agent', 'agent_no_tests'),
+    main.add_argument('--llm-mode',
+        choices=('default', 'no_ffi', 'agent', 'agent_sim_no_tests'),
         default='default',
         help='which style of LLM-based rewriting to use')
 
     safety_loop = sub.add_parser('safety-loop')
     safety_loop.add_argument('--c-code', default='c_code')
     safety_loop.add_argument('node', nargs='?', default='current')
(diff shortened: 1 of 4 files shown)


fw-immunant (approved):
Generally LGTM. One pattern I would maybe suggest changing is the definition of top-level constants that are format strings - I would instead expose these as functions that perform formatting internally so there's no question about how they need to be used by consumers. But that's just a style nit, not a blocking concern.

I also looked at the `related_decls` [test failure](https://github.com/GaloisInc/Tractor-Crisp/actions/runs/24580980492/job/71878467064?pr=104). I'm not sure how it's possible, and it probably deserves some investigation. But it definitely shouldn't have anything to do with the changes that this PR makes...

spernsteiner:
OK, I'll merge this now, but we should definitely investigate the `related_decls` error later.  We were getting similarly incomprehensible errors from those tests at one point in the past, but https://github.com/GaloisInc/Tractor-Crisp/pull/82/changes/fe0c1a311bf9a3e239a107fef6faa7baf44c878f seemed to fix it that time.

PR #106, opened 2026-04-21
tools: update to rust-analyzer 0.0.329

spernsteiner:
This should fix the CI errors we've been seeing with the related_decls tests.  I gave the CI logs to Codex and it pointed out that the errors started after Rust 1.95.0 was released and our CI automatically switched over to the new version.  Since the test failures involved a `const` that was used only within the arguments of an `eprintln!` macro call, I figured maybe something had changed in the definition of `eprintln!` causing rust-analyzer to have trouble with it.  I had Codex upgrade the tools to `ra_ap_*` 0.0.329 (from 0.0.301), and the tests now pass.  I don't know the actual root cause.

The one notable change I saw in the diff is the replacement of some `x.ty(db)` calls (to get the type of `x`) with `x.ty(db).to_type(db)`.  rust-analyzer upstream has a new type representation `ra_ap_hir::TypeNs`, and some `ty` methods (e.g. `Field::ty`) now return this, while others (e.g. `Param::ty`) return the previous `ra_ap_hir::Type`.  Conversion is allowed only in one direction (`TypeNs` -> `Type`), so I think the current approach of converting everything to `Type` is the best we can do for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,9 @@
 # Debian trixie (13) has gcc-13.
 FROM docker.io/rust:trixie AS tractor-crisp-user
 
-# rust-analyzer (required by hayroll)'s deps require Rust 1.89
-RUN rustup default 1.90.0
-RUN rustup +1.90.0 component add rustfmt
+# rust-analyzer 0.0.329 (used by tools/*) requires Rust 1.93 at minimum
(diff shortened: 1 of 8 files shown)


fw-immunant (approved):
Changes look reasonable.

PR #112, opened 2026-05-01
Fix chunked encoding error

souryadey:
While running some experiments, I ran into `requests.exceptions.ChunkedEncodingError: Response ended prematurely`. Some digging suggested that this is during the LLM rewrite calls and is likely due to dropped connections / server issues. This MR attempts to fix it by wrapping the requests response handling in a `try` block, and handling `requests.exceptions`s gracefully via `except`. I ran a couple of experiments with the changed code and didn't run into these errors any more.

### DIFF
--- a/crisp/llm.py
+++ b/crisp/llm.py
@@ -55,9 +55,8 @@ def sse_events(resp):
     except (
         requests.exceptions.ChunkedEncodingError,
         requests.exceptions.ConnectionError
-    ):
-        if acc:
-            yield bytes(acc)
+    ) as e:
+        raise CrispError('Could not parse LLM API request response') from e
 
     # If there are bytes remaining in `acc`, don't yield them.  "If the file
     # ends in the middle of an event, before the final empty line, the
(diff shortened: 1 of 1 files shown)


crisp/llm.py
```diff
@@ -29,26 +29,35 @@ def sse_events(resp):
...
+            if value.startswith(b' '):
+                value = value[1:]
+            acc.extend(value)
+
+    except (
+        requests.exceptions.ChunkedEncodingError,
+        requests.exceptions.ConnectionError
+    ):
+        if acc:
             yield bytes(acc)
```
spernsteiner: I don't think it's correct to suppress the error and yield a partial `acc` here.  This should throw a `CrispError` instead, which signals to `safety_loop_common` that something has failed and it should retry the current operation.  (You may need to add similar logic in your GEPA code.)

souryadey: Done.

spernsteiner (approved):

PR #113, opened 2026-05-01
Fix too many open files

souryadey:
While running some experiments, I ran into `OSError: Too many open files`. Some digging suggested that this is due to `open()` calls, `subprocess.Popen()` calls and responses not being closed, leading to too many open files. This can be gotten around by setting `ulimit -Sn` to a higher value (limit is the value of `ulimit -Hn`) before running the experiments, but this is a workaround instead of an actual fix. Hopefully the changes in this MR – `with open`, `p.stdout.close()` and `resp.close()` – are more of an actual fix.

I ran a couple of experiments with the changed code and a higher `ulimit` and didn't run into these errors any more.

### DIFF
--- a/crisp/config.py
+++ b/crisp/config.py
@@ -32,10 +32,11 @@ class ConfigBase:
     def from_toml_file(cls, f, **kwargs):
         if isinstance(f, str):
             path = f
-            f = open(f, 'r')
+            with open(f, 'r') as f_:
+                d = toml.load(f_)
         else:
             path = f.name
-        d = toml.load(f)
+            d = toml.load(f)
         return cls.from_dict(d, path, **kwargs)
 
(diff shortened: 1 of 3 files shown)


spernsteiner (commented):
I don't understand why this is needed.  Shouldn't `p`/`f`/`resp` be closed automatically when they go out of scope?

souryadey:
> I don't understand why this is needed. Shouldn't `p`/`f`/`resp` be closed automatically when they go out of scope?

I don't think that's guaranteed. From what I gleaned by digging online, files that are opened without a context manager (i.e. `open()` as opposed to `with open()`) may not be automatically closed when the code containing the `open()` statement goes out of scope.

Subprocess.Popen() and request responses with `stream = True` also may not be closed immediately.

PR #114, opened 2026-05-01
tools: add unsafe static item tracking to find_unsafe

### LINKED ISSUE #100: Update the find_unsafe tool to cover static mut globals
In our first libexpat experiment at https://github.com/immunant/libexpat/tree/experimental/tractor/ahomescu/crisp1, the agent completely ignored `static mut` globals like `utf8_encoding_ns` from https://github.com/immunant/libexpat/blob/experimental/tractor/ahomescu/crisp1/translated_rust/src/lib/xmltok.rs#L6476 because it was not included in the prompt or counted by the unsafe tool.

We should extend the tool to produce a count of unsafe globals like
```
static mut foo: Type = unsafe { value };
```

Crocodoctopus:


Crocodoctopus:
I've implemented tracking for static items containing unsafe, but I've only tested it via running the tool directly against a few test files I wrote. I assume this isn't enough to integrate the changes into the project at large?

ahomescu:
It's used to refactor https://github.com/immunant/libexpat, you could try to run it on that. Does the tool not have its own tests though?

ahomescu (changes requested):
Ready to go as soon as we have some confirmation of testing.

tools/find_unsafe/src/main.rs
```diff
@@ -46,20 +43,25 @@ fn fn_is_exported(f: &ItemFn) -> bool {
...
-#[derive(Clone, Debug, Default)]
-#[derive(Serialize)]
+#[derive(Clone, Debug, Default, Serialize)]
 struct Output {
     /// Functions that are not accessible from other compilation units and are also unsafe.
     internal_unsafe_fns: Vec<String>,
     /// Functions that contain an unsafe block.
     fns_containing_unsafe: HashSet<String>,
+    /// Static that contain an unsafe block.
+    statics_containing_unsafe: HashSet<String>,
```
ahomescu, line 63: It might be useful to also have a separate counter for `static mut`s with or without `unsafe` initializers. For example `static mut FOO: i32 = 0;` is still unsafe to read or write.

Crocodoctopus:
> It's used to refactor https://github.com/immunant/libexpat, you could try to run it on that. Does the tool not have its own tests though?

The tool itself has very minimal testing, doesn't look like anything related to verifying if its counting correctly. I haven't actually run the crisp loop yet, only the tool itself. Looking into doing that now.

Also, I was thinking, shouldn't the tool output the fully qualified name? There are probably not many collisions in practice, but they are possible.

tools/find_unsafe/src/main.rs
```diff
@@ -79,15 +83,32 @@ impl<'ast> Visit<'ast> for Visitor {
         self.current_fn = old;
     }
 
+    fn visit_item_static(&mut self, item_static: &'ast ItemStatic) {
+        let name = item_static.ident.to_string();
+        if static_is_exported(item_static) {
+            return;
+        }
```
spernsteiner: Unlike with functions, I think we do want to count unsafe code in exported statics.  Unlike with functions, we can't really split these up into an unsafe FFI wrapper and a safe core.  Also, actual public-facing statics should be extremely rare (if we see a `#[no_mangle]` static, it's more likely that it's just being exported from one C compilation unit to another, rather than being exported for library users to access).

tools/find_unsafe/src/main.rs
```diff
@@ -79,15 +83,32 @@ impl<'ast> Visit<'ast> for Visitor {
...
+
+        if matches!(item_static.mutability, StaticMutability::Mut(_)) {
+            self.out.mutable_statics.insert(name.clone());
+        }
+
+        let old = self.current_static.replace(name);
+        visit::visit_item_static(self, item_static);
+        self.current_static = old;
```
spernsteiner: This should maybe clear `current_fn` for the duration of the recursive visit, to avoid double counting in cases like this:
```Rust
fn f() {
    static S: i32 = unsafe { 0 };
}
```

spernsteiner (commented):
Please don't run rustfmt on things in this repo—IME it usually makes the code worse instead of better

spernsteiner:
> shouldn't the tool output the fully qualified name?

Does it not do so currently?  I guess we mostly just count the amount of unsafe rather than looking at the actual names in the list, but I thought it was emitting qualified paths like our other tools do.

spernsteiner:
> The tool itself has very minimal testing, doesn't look like anything related to verifying if its counting correctly

If you feel like adding a proper test case for this tool, that would certainly be nice to have

Crocodoctopus:
> > shouldn't the tool output the fully qualified name?
> 
> Does it not do so currently? I guess we mostly just count the amount of unsafe rather than looking at the actual names in the list, but I thought it was emitting qualified paths like our other tools do.

```
fn f() { unsafe { } }

fn x() {
    fn f() { unsafe { } }
}

fn y() {
    fn f() { unsafe { } }
}
```
Given something like this, the tool outputs `fns_containing_unsafe":["f"]`. Only if all of these became safe would we have made progress as far as the tool is concerned. Switching from set -> list will at least yield "f" 3 times, which sounds like its closer to what you want, but "f" is still ambiguous.

tools/find_unsafe/src/main.rs
Crocodoctopus: I've updated it, only the most recent fn/static is considered when encountering an unsafe block.

Crocodoctopus:
Integration into the crisp loop will require rewriting some prompts, particularly [`crisp/workflow.py:102`](https://github.com/GaloisInc/Tractor-Crisp/blob/3bf6c8c23ed61af487a7cf4558a753eea611ef7e/crisp/workflow.py#L102). I'm no prompt engineer, any suggestions? I guess I can always just keep trying things until output improves.

ahomescu:
> Integration into the crisp loop will require rewriting some prompts

I thought we'd land this first by itself, but I guess that works too. @spernsteiner Do you have a preference here?

Crocodoctopus:
> > Integration into the crisp loop will require rewriting some prompts
> 
> I thought we'd land this first by itself, but I guess that works too. @spernsteiner Do you have a preference here?

I haven't done any work on integrating it yet, so if you want, we can land now.

tools/find_unsafe/src/main.rs
```diff
@@ -53,18 +56,29 @@ struct Output {
     /// Functions that are not accessible from other compilation units and are also unsafe.
     internal_unsafe_fns: Vec<String>,
     /// Functions that contain an unsafe block.
-    fns_containing_unsafe: HashSet<String>,
+    fns_containing_unsafe: Vec<String>,
```
spernsteiner: Won't this now end up with multiple entries for the same function, if you have code like this?
```Rust
fn f() {
    unsafe { 1 };
    unsafe { 2 };
}
```
And similarly for the new `statics_containing_unsafe` below.

If ordering/determinism is important, we could use `indexmap::IndexSet` instead of `HashSet` here.

tools/find_unsafe/src/test.rs
```diff
@@ -0,0 +1,63 @@
...
+static B: () = {
+    static C: () = unsafe { };
+};
+
+// Mark D, not E
+static D: () =
+    unsafe { static E: () = { }; };
+
+///////////
+// `static mut X` is treated as `static X = unsafe { .. }` by the tool.
```
spernsteiner: Inaccurate comment.  `static mut X` goes in `mutable_statics`, while `static X = unsafe { ... }` goes in `statics_containing_unsafe`.

tools/find_unsafe/src/test.rs
```diff
@@ -0,0 +1,63 @@
...
+unsafe fn f() {
+    fn g() { unsafe { } }
+    static F: () = unsafe { };
+    static mut _F: () = { };
+}
+
+// Mark h twice
+unsafe fn h() {
+    unsafe { }
+}
```
spernsteiner: Might be worth noting that these are two different kinds of marks, `internal_unsafe_fns` vs `fns_containing_unsafe` (and similarly for `_G` below).

spernsteiner:
I agree it's fine to land this without CRISP integration.  These changes are just adding new outputs, so there should be no breakage for CRISP's existing uses of the tool.

tools/find_unsafe/src/main.rs
Crocodoctopus: Moved it back to HashSet. Didn't think of this case, was thinking about situations like:
```
fn f() {
  unsafe { }
  fn f() {
  unsafe { }
  }
}
```
Which is a bit contrived, but still possible. I agree vec isn't a good choice either, we should support full names properly, eventually at least.

tools/find_unsafe/src/test.rs
Crocodoctopus: Comment removed. Its not a requirement that the tool do this anyways, and may change.

tools/find_unsafe/src/test.rs
Crocodoctopus: Fixed, tests got a bit reordered though.

spernsteiner (approved):

ahomescu:
LGTM!

PR #116, opened 2026-05-05
GEPA

souryadey:
Add code and results for GEPA prompt optimization, integrated with the CRISP workflow.

- `crisp/gepa_po.py` is the main module containing classes and functions related to GEPA. Consequently, `crisp/workflow.py` is updated with a new step `llm_gepa`, and `crisp/llm.py` has its corresponding code to query LLMs using separate system and user messages.
- `scripts/run_gepa.py` is the script to run GEPA and evaluate its performance.
- `gepa_artifacts/` contains found prompts and results from GEPA.
- README is updated with how to run GEPA, and the repo root contains the associated script `gepa_setup_initial.sh`.

Requires new dependencies `gepa` (obviously 😄), `litellm`, and `pandas`. All of these are marked as optional, to be used when running GEPA.


converted_rust_projects/README.md
```diff
@@ -0,0 +1 @@
+Each zip file in this folder is named as `c2rust_<dataset>.zip`, and contains Rust projects. These were obtained via running C2Rust in bulk on the `<dataset>` of C projects. Unzipping each zip file will lead to one folder for each successfully converted Rust project from `<dataset>`, and a `results_transpilation.csv` documenting transpilation success.
```
spernsteiner: Nothing uses the `converted_rust_projects` directory any more, so it can be removed.

crisp/llm.py
```diff
@@ -253,27 +254,36 @@ def run_rewrite(
...
+            output_instructions=file_formatter.get_output_instructions(),
+            output_instructions_lowercase=file_formatter.get_output_instructions_lowercase(),
+            **format_kwargs,
+        )
+        req_messages = [
+            {'role': 'user', 'content': prompt},
+        ]
+
+    else:
+        prompt_without_files = prompt_fmt
```
spernsteiner: For consistency, can you have this call `.format(...)` with the same arguments as above (except for `input_files`)?  Then we can toggle `separate_system_prompt` on or off without having to make any changes to `prompt_fmt`.

.gitignore
```diff
@@ -10,3 +10,13 @@ target/
 
 /crisp-bwrap.toml
 /bwrap-sandbox/
+
+# GEPA-related ignored files
+gepa_setup_initial.sh
+gepa_setup_rerun.sh
+scripts/expt_gepa.py
+scripts/expt_gepa_via_dspy.py
```
spernsteiner: Normally `.gitignore` is for files that are produced by the project itself (e.g. build artifacts), since those need to be ignored on every developer's machine.  If this is something that will only exist on your machine, you can put them in `.git/info/exclude` instead.

.gitignore
```diff
@@ -10,3 +10,13 @@ target/
...
+
+# Ignore scripts that have been subsumed by CRISP workflow
+scripts/run_c2rust_and_get_datasets.py
+scripts/llm_unsafec2rust_to_saferust.py
```
spernsteiner: These scripts no longer exist, so there should be no need to list them in `.gitignore`.

pyproject.toml
```diff
@@ -5,6 +5,9 @@ requires-python = "==3.14.*" # pygit2 doesn't have prebuilts for 3.15 yet.
 dependencies = [
     "cbor>=1.0.0",
     "docker>=7.1.0",
+    "gepa>=0.1.1",
+    "litellm>=1.83.14",
+    "pandas>=3.0.2",
```
spernsteiner: Is there an easy way to make these dependencies optional?  AIUI we will mostly be running GEPA very infrequently (e.g. only when upgrading to a new model), so they won't be needed most of the time.  And of course I'm thinking of the recent LiteLLM supply chain attack.

scripts/run_gepa.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+
+    for prompt_name in [
+        # insert prompt names for evaluation here
+    ]:
+        evaluate_gepa_found_prompt(
+            dataset_path = Path(__file__).resolve().parent.parent / 'Test-Corpus/Public-Tests/B01_organic',
+            optimized_prompt_folder = Path(__file__).resolve().parent.parent / f'gepa_artifacts/{prompt_name}',
+            model = 'gpt-5.5'
+        )
```
spernsteiner, line 30: What's the connection between `seed_prompt_2` in the first call and `{prompt_name}` in the second?  I'm not sure what's supposed to be hardcoded vs customizable here.

crisp/gepa_po.py
```diff
@@ -0,0 +1,417 @@
...
+    separator: str = '\n\n'
+) -> str:
+    """
+    Get the contents of a node's files whose paths match given patterns.
+
+    Inputs:
+    - node: The node whose files to look at.
+    - patterns: List of string path patterns. Expected to contain wildcards. Any file in `node.files` which matches any pattern *from the right* will be considered.
+    - mvir: The MVIR in consideration.
+    - separator: After getting matching files, extract their body contents and concatenate the strings using this separator.
```
spernsteiner: With the default separator of `\n\n`, won't it be hard for the model to tell where files begin and end, or to see which file is which?

I think you could instead call `get_file_formatter('xml').emit_files(mvir, node, glob_filter=patterns)`, which will combine all the matching files into a string with `<file name="foo.rs">...</file>` delimiters around each one.  (This is the same logic that's used internally within `llm.run_rewrite` to combine many files into a single prompt.)

crisp/gepa_po.py
```diff
@@ -0,0 +1,417 @@
...
+            )
+
+        # Check for un-safety
+        unsafe_results = workflow.find_unsafe_op(n_llm_output_code).body_json()
+        internal_unsafe_fns = []
+        fns_containing_unsafe = []
+        for result in unsafe_results.values():
+            internal_unsafe_fns.extend(result['internal_unsafe_fns'])
+            fns_containing_unsafe.extend(result['fns_containing_unsafe'])
+        total_unsafe = len(internal_unsafe_fns) + len(fns_containing_unsafe)
```
spernsteiner: This should use `workflow.count_unsafe(...)` to get the total count instead of adding it up manually.  We are currently looking at changing how unsafe is counted (in order to improve the results on libexpat), and it's good to have that logic defined in one place so we don't have multiple versions that may get out of sync.  `find_unsafe_op` caches its results automatically, so after calling `find_unsafe_op` above, calling `count_unsafe` is very cheap.

crisp/gepa_po.py
```diff
@@ -0,0 +1,417 @@
...
+        scores = []
+        trajectories = [] if capture_traces else None
+
+        for task in batch:
+
+            n_c_code_id = parse_node_id_arg(task['workflow'].mvir, 'c_code')
+            n_c_code = task['workflow'].mvir.node(n_c_code_id)
+
+            n_llm_input_code_id = parse_node_id_arg(task['workflow'].mvir, 'current')
+            n_llm_input_code = task['workflow'].mvir.node(n_llm_input_code_id)
```
spernsteiner: Please note in a comment that this code is assuming (IIRC) that `current` is the non-rewritten, unsafe C2Rust output, as produced by running `test_eval.py` with `LLM_SAFETY_TRIES=1` on each example project.  There should probably also be a comment in the main `run_gepa.py` script to let the user know they need to set things up this way.

pyproject.toml
```diff
@@ -5,6 +5,9 @@ requires-python = "==3.14.*" # pygit2 doesn't have prebuilts for 3.15 yet.
 dependencies = [
     "cbor>=1.0.0",
     "docker>=7.1.0",
+    "gepa>=0.1.1",
+    "litellm>=1.83.14",
```
spernsteiner: Actually, it looks like we're no longer using `litellm` in any of these scripts, so please remove it from the deps if that's correct.

spernsteiner (commented):
Looks good—all these comments are relatively minor

spernsteiner:
If you rebase onto current `main`, the build-tools CI error should go away.  It might also be good to squash down some/all of the history here, since IIRC there were several datasets that were at one point added and later removed (but are still present in the history and taking up space)

converted_rust_projects/README.md
souryadey: Done. Changes will be reflected with the next push.

.gitignore
souryadey: Good suggestion. Done. Changes will be reflected with the next push.

.gitignore
souryadey: Moved to `.git/info/exclude`. Changes will be reflected with the next push.

pyproject.toml
souryadey: Yes there is, via [optional dependencies](https://pydevtools.com/handbook/explanation/what-are-optional-dependencies-and-dependency-groups/). I've added an optional dependency group for gepa. Changes will be reflected with the next push.

crisp/gepa_po.py
souryadey: Ah, in that case, sure. Done, changes will be reflected with the next push.

pyproject.toml
souryadey: gepa needs litellm to work. I originally tried to run gepa without explicitly installing litellm and it failed. Anyway, I've moved litellm to optional dependencies (see [here](https://github.com/GaloisInc/Tractor-Crisp/pull/116#discussion_r3276894778).)

crisp/gepa_po.py
souryadey: It's actually `LLM_SAFETY_TRIES=0`,  but yes, that's a good point about better documentation. In fact, there should be clear documentation on how to run GEPA. I'll put in a comment here as you suggested, and also update `gepa_artifacts/README` to include more details.

crisp/llm.py
souryadey: Because GEPA optimizes the prompt string itself, it does not work very well when the prompt has `{...}` blocks to be filled in via `.format()`. In other words, the best way to GEPA-optimize prompts is to just give it the entire prompt. Since it's usually the system prompt that's being optimized, the expectation is that the system prompt  doesn't contain any `{...}` blocks since any user input blocks can become part of the user prompt. This is why the following lines are:
```python
req_messages = [
    {'role': 'system', 'content': prompt_without_files},
    {'role': 'user', 'content': input_files_str}
]
```

An example of the system prompt I'm starting from for GEPA is [here](https://github.com/GaloisInc/Tractor-Crisp/blob/f2383fc71fe32e98fe0def1fb6b9ec70cdd1e9e7/gepa_artifacts/seed_prompt_2/prompt.txt).

So, in a nutshell, I'm not sure if we should edit this code to include `prompt_fmt.format(...)`, since `prompt_fmt` is a system prompt with no `{...}` blocks to be filled in.

spernsteiner: If `prompt_fmt` has no `{...}` placeholders, then calling `.format()` will have no effect, so this shouldn't cause any problems for the GEPA case.  But adding it would let us try turning on `separate_system_prompt` in production (when `prompt_fmt` does have placeholders that need to be filled in), and overall would be less surprising (`separate_system_prompt=True` just sends the system prompt separately, instead of sending the system prompt separately and also turning off formatting).

scripts/run_gepa.py
souryadey: The first block of code calling `do_gepa()` is actually running GEPA optimization using `seed_prompt_2` as the starting point.

The second block of code calling `evaluate_gepa_found_prompt()` will run the GEPA evaluation functions on any prompt. The elements in the `for prompt_name in []` list can be any prompts, whether GEPA-found or seed. In fact, we expect to include both GEPA-found prompts and seed prompts in the list so that we can quantify improvement.

I've clarified all this via comments, and also renamed `evaluate_gepa_found_prompt()` to `run_gepa_eval_on_prompt()`. Changes will be reflected in the next push.

crisp/llm.py
souryadey: Sure, I can do that.

Done. Now, the only thing that's different between the `if not separate_system_prompt` and the `else` block is the `req_messages` variable.

Changes will be reflected in the next push.

crisp/gepa_po.py
souryadey: Added documentation, and a setup script.

Changes will be available in the next push.

crisp/gepa_po.py
souryadey: Great suggestion! I have implemented this. Changes will be available in the next push.

souryadey:
@spernsteiner : Thank you for your careful and detailed review! I think I've addressed all the comments. I've also pulled in the latest `main`, which makes all CI checks pass.

spernsteiner:
The final version of this branch looks good, but please squash the history down to a single commit.  The current history is pretty messy (100+ commits including 11 different merges, adding things only to later remove them, including multi-MB datasets).

souryadey:
So to clarify, you mean that I should go ahead and merge this using the "Squash and merge" option, correct?

spernsteiner (approved):
>you mean that I should go ahead and merge this using the "Squash and merge" option, correct?

Ah, yes, github's "squash and merge" should work for this.

PR #117, opened 2026-05-08
tools: add unsafe macro tracking to find_unsafe

### LINKED ISSUE #101: Update the find_unsafe tool to cover macros
In our first libexpat experiment at https://github.com/immunant/libexpat/tree/experimental/tractor/ahomescu/crisp1, the agent kept moving unsafe code into macros. The `find_unsafe` tool should probably count those too and include them in its output.

Crocodoctopus:
Adds token based unsafe finding to the find_unsafe tool, as well as a small test.

### DIFF
--- a/tools/find_unsafe/Cargo.toml
+++ b/tools/find_unsafe/Cargo.toml
@@ -9,3 +9,4 @@ syn.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 clap.workspace = true
+proc-macro2.workspace = true
--- a/tools/find_unsafe/src/main.rs
+++ b/tools/find_unsafe/src/main.rs
@@ -2,10 +2,10 @@ use std::collections::{HashMap, HashSet};
 use std::fs;
 use std::io::{self, Read};
 use std::path::{self, PathBuf};
-use ciborium;
 use clap::Parser;
+use proc_macro2::{TokenStream, TokenTree};
 use serde::Serialize;
-use syn::{self, Attribute, ExprUnsafe, Ident, ItemFn, ItemStatic, Meta, Path, StaticMutability};
+use syn::{self, Attribute, ExprUnsafe, Ident, ItemFn, ItemMacro, ItemStatic, Macro, Meta, Path, StaticMutability};
 use syn::visit::{self, Visit};
 
 // Include test files to ensure they compile.
@@ -13,6 +13,8 @@ use syn::visit::{self, Visit};
 mod test_funcs;
 #[allow(warnings)]
 mod test_statics;
+#[allow(warnings)]
+mod test_macros;
 
 fn is_link_attr(attr: &Attribute) -> bool {
     is_link_attr_meta(&attr.meta)
@@ -51,6 +53,17 @@ fn fn_is_exported(f: &ItemFn) -> bool {
     f.attrs.iter().any(is_link_attr)
 }
 
+fn token_stream_contains_unsafe(tokens: TokenStream) -> bool {
+    for token in tokens {
+        match token {
+            TokenTree::Ident(ident) if ident == "unsafe" => return true,
+            TokenTree::Group(group) if token_stream_contains_unsafe(group.stream()) => return true,
+            _ => { continue }
+        }
+    }
+    false
+}
+
 
 #[derive(Clone, Debug, Default)]
 #[derive(Serialize)]
@@ -63,6 +76,10 @@ struct Output {
     statics_containing_unsafe: HashSet<String>,
     /// Static that are mutable, regardless of unsafe.
     mutable_statics: HashSet<String>,
+    /// Macro invocations that contain an unsafe token, and don't belong to an item (ex global invocations).
+    global_macro_invocations_containing_unsafe: HashSet<String>,
+    /// Macro definitions (macro_rules!) that contain an unsafe token.
+    macro_definitions_containing_unsafe: HashSet<String>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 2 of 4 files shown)


tools/find_unsafe/src/main.rs
```diff
@@ -85,6 +102,36 @@ impl<'ast> Visit<'ast> for Visitor {
...
+            .chain(segments
+                .iter()
+                .map(|seg| seg.ident.to_string()))
+            .collect::<Vec<_>>()
+            .join("::");
+
+
+        if token_stream_contains_unsafe(mac.tokens.clone()) {
+            self.out.macro_invocations_containing_unsafe.insert(name);
+        }
```
spernsteiner, line 171: `unsafe` in macro invocations should be attributed to the item that contains it.  E.g. in `fn f() { m!(unsafe { ... }) }`, we should report that `f` contains some unsafe.  (One thing I'm looking at implementing soon is using the detailed output from `find_unsafe` to find a function containing unsafe code to suggest as a starting point when running the agent.)

tools/find_unsafe/src/main.rs
```diff
@@ -85,6 +102,36 @@ impl<'ast> Visit<'ast> for Visitor {
         }
         visit::visit_expr_unsafe(self, x);
     }
+
+    fn visit_item_macro(&mut self, item_mac: &'ast syn::ItemMacro) {
+        let Some(name) = item_mac.ident.as_ref().map(|i| i.to_string()) else {
```
spernsteiner, line 140: Maybe worth noting that the `ident` here is not the name of the macro we're calling, but rather the `foo` in `macro_rules! foo { ... }`

tools/find_unsafe/src/main.rs
```diff
@@ -85,6 +102,36 @@ impl<'ast> Visit<'ast> for Visitor {
...
+            // This is an invocation, pass it along.
+            visit::visit_item_macro(self, item_mac);
+            return;
```
spernsteiner, line 143: We should check the macro arguments for `unsafe` here, just like we do for expr macros.  (Really we should do this for macro invocations in all positions, but I think expr, stmt, and item are the most important / most likely to be used by the agent.)  I'm not sure what to attribute the unsafety to, since we don't know the name of the item (if any) generated by this macro invocation.  I could imagine either providing a file and line number from the span of the invocation, or else reporting the macro path that's being called (e.g. "we found unsafe in the arguments of a call to macro `foo::bar!`") and let the agent grep to figure out which call site is the relevant one.

tools/find_unsafe/src/main.rs
Crocodoctopus: Done. `macro_invocations_containing_unsafe` has been repurposed to hold global invocations only.

tools/find_unsafe/src/main.rs
Crocodoctopus: Done, Left better comments.

tools/find_unsafe/src/main.rs
Crocodoctopus: I believe the feature you want is already implemented. `visit_macro` generically matches invocations, and `macro_rules` declarations. `visit_item_macro` overrides this just for the `macro_rules`, since that one is special. All macro invocations types are currently supported, and any unsafe in them is attributed to the item they're in (currently, their immediate function or static).

I can probably add a line number. Would we want to even record the name at all? The line number would fully (well, almost) disambiguate, though knowing immediately what function we're in might still be useful information even if it could be calculated after the fact.

Thoughts?

spernsteiner: Ah, it didn't occur to me that `visit_item_macro` would dispatch to `visit_macro`.

For item macros (not inside a function or static), either name or file/line number should be fine.  The thing I want to avoid here is having the tool just say "there are 3 item macro invocations that contain unsafe code" without giving the agent any clue as to where among the 100k lines of the codebase those invocations actually are.

Crocodoctopus:
As is, the current behavior is: 
- global invocation macros record the **name of the macro they're invoking**
- non-global invocation macros record the **name of the item they're invoked in** (function or static name)
- "macro_rules!" record the macro name in all contexts

The tool produces:
```
{
  "src/test_macros.rs":{
    "internal_unsafe_fns":[],
    "fns_containing_unsafe":["demo"],
    "statics_containing_unsafe":[],
    "mutable_statics":[],
    "global_macro_invocations_containing_unsafe" ["unsafe_within_invocation2"],
    "macro_definitions_containing_unsafe":["unsafe_ItemMacro","unsafe_TypeMacro","unsafe_ExprMacro","false_positive","unsafe_StmtMacro"]
  }
}
```
for the contents within `src/test_macros.rs`. If this behavior is correct then we should be about ready to merge.

Also, I should mention the tool does mark this as unsafe:
```
macro_rules! false_positive {
    ($unsafe:tt) => { $unsafe }
}
```
I can't imagine this will matter, but any single token that can be "unsafe" will be picked up by the tool.

tools/find_unsafe/src/main.rs
```diff
@@ -63,6 +76,9 @@ struct Output {
     statics_containing_unsafe: HashSet<String>,
     /// Static that are mutable, regardless of unsafe.
     mutable_statics: HashSet<String>,
+    /// Macros that contain an unsafe token.
+    global_macro_invocations_containing_unsafe: HashSet<String>,
```
spernsteiner:
```suggestion
    /// Names of global/item macros that are called on an unsafe token.
    global_macro_invocations_containing_unsafe: HashSet<String>,
```

tools/find_unsafe/Cargo.toml
```diff
@@ -9,3 +9,4 @@ syn.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 clap.workspace = true
+proc-macro2 = "1"
```
spernsteiner:
```suggestion
proc-macro2.workspace = true
```
All dependencies should go through the workspace manifest so it's easy to keep the versions and features consistent.

spernsteiner (approved):

tools/find_unsafe/src/main.rs
Crocodoctopus: Fixed the comment. Also gave a comment to the other undocumented container.

tools/find_unsafe/Cargo.toml
Crocodoctopus: Just replacing line 12 with `proc-macro2.workspace = true` is enough?

spernsteiner: Depends on whether we have the `proc-macro2` dependency already listed in the workspace config (I would assume we do, but I haven't checked)

Crocodoctopus: Looks like it is, and it builds, so we're good.

ahomescu (approved):

PR #118, opened 2026-05-15
add tools/find_unsafe2

spernsteiner:
This adds a new `find_unsafe2` tool that provides a fine-grained breakdown of unsafe operations in each function.  There's also an accompanying `check_unsafe2` tool that compares the current state of the code to the previous `find_unsafe2` JSON output and reports an error if any new unsafety has been added (e.g. adding a new function that contains unsafe, or adding a third raw pointer dereference to a function that previous had only two).  The intent is to use this as a stricter guardrail for agent-based rewriting, to prevent the agent from "cheating" using macros or unsound "safe" helper functions.

The new tools are implemented using `rustc_public` (a.k.a. "Stable MIR"—it is not yet stable, but the intent is to stabilize it relatively soon).  This means the tools are tied to a specific Rust nightly, which so far we've been trying to avoid due to the maintenance burden this caused on previous projects.  The hope is that using `rustc_public` will keep the maintenance burden lower because many compiler-internal changes and refactorings will remain hidden behind the `rustc_public` interface so that we don't have to modify our code to account for them.

These tools currently report:
* Raw pointer dereferences
* Calls to unsafe functions
* Accesses of mutable statics
* Union field accesses

They currently don't report:
* Implementations of unsafe traits
* FFI import declarations
* Unsafe attributes on functions

See the comments on `Outputs` and `FunctionOutputs` for details.

In addition, these tools report specifically which FFI imports are used in each function, so the agent must use safe Rust libraries in its rewrites instead of introducing new libc calls that will need to be patched out later.  In the future we can add more progress heuristics like this, such as counting the number of raw pointer types or the number of `unsafe fn`s defined in the codebase.

Some other miscellaneous notes:
* This doesn't have carve-outs for FFI entry points, which will be necessary before it can replace the current `find_unsafe` within the CRISP loop.
* The tools compute callee and struct-field usage info as part of unsafe checking, which we might be able to use for callgraph analysis, in place of the rust-analyzer-based `related_decls` tool.
* This branch includes cargo wrappers `cargo find-unsafe2` and `cargo check-unsafe2` that run the tools across an entire Cargo workspace, so it should be usable on projects with multiple build artifacts.


randomPoison (approved):
Looks good to me, glad this wasn't too difficult to put together. Hopefully rustc_public stays stable enough that this doesn't end up difficult to maintain 🤞

PR #120, opened 2026-05-21
test_eval: add clar suite support & suite selector

Crocodoctopus:
Preliminary (-ish) PR for adding clar support. 

In addition to that, it introduces a new `--suites` flag and folds some other flags into it (ex. `--no-runtests` is now `--suites none`). Still iffy about backwards compatibility, we'll see what CI has to say.

### DIFF
--- a/scripts/test_eval.py
+++ b/scripts/test_eval.py
@@ -8,6 +8,7 @@ import json
 import os
 import shlex
 import subprocess
+import sys
 import tempfile
 import toml
 
@@ -16,22 +17,48 @@ import toml
 class Args:
     project_dir: Path
     extra_test_dirs: list[Path]
-    runtests: bool
     git: bool
+    relax_implicit_none: bool
+    cmake_preset: str | None
+    suite: str
     main_args: list[str]
+
+
+def detect_suites(case_dir: Path) -> list[str]:
+    # Keep suite detection limited to paths this script itself assumes later.
+    has_runtests = (case_dir / "test_vectors").is_dir()
+    has_clar = (
+        (case_dir / "tests").is_dir()
+        and (case_dir / "tests/CMakeLists.txt").is_file()
+    )
+
+    suites = []
+    if has_runtests:
+        suites.append("runtests")
+    if has_clar:
+        suites.append("clar")
+    return suites
 
 
 def parse_args() -> Args:
     ap = argparse.ArgumentParser()
     ap.add_argument("project_dir", type=Path)
+    ap.add_argument("--suite",
+        choices=("auto", "runtests", "clar", 'none'),
+        default="auto",
+        help="which test suite to use: auto-detect one, force a specific suite, or disable tests with 'none'")
+    ap.add_argument("--relax-implicit-none",
+        action='store_true', dest='relax_implicit_none',
+        help="when --suite=auto and no usable suites are detected, continue with suite='none' instead of raising an error")
+    ap.add_argument("--cmake-preset",
+        dest='cmake_preset',
+        help="name of CMake build preset to use when CMakePresets.json has multiple buildPresets")
     ap.add_argument('--extra-test-dir',
         type=Path, action='append', default=[], dest='extra_test_dirs',
         help='run extra test vectors from this directory')
-    ap.add_argument("--no-runtests", action='store_false', dest='runtests',
-        help="don't import `runtests` scripts or use them in `test_command`")
     ap.add_argument("--no-git", action='store_false', dest='git',
         help="don't require the project dir to be inside a git repository "
-            '(implies --no-runtests)')
+            '(implies runtests suite is unavailable)')
     ap.add_argument("main_args", nargs='*',
         help='extra arguments to pass to `crisp main`')
     return Args(**ap.parse_args().__dict__)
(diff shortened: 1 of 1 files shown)


scripts/test_eval.py
```diff
@@ -156,16 +172,43 @@ def relpath(path: Path, start: Path) -> Path:
 def main():
     args = parse_args()
 
-    if not args.git:
-        args.runtests = False
-    if not args.runtests:
-        assert len(args.extra_test_dirs) == 0, \
-            '--extra-test-dirs is not supported with --no-runtests'
```
spernsteiner, line 163: This check should be reinstated, since it's still the case that only `suite == "runtests"` supports `extra_test_dirs` (which is fine—`extra_test_dirs` exists mainly due to a quirk of the P01 layout, which uses `runtests`).

spernsteiner (commented):
CI does test this (and seems to be passing), but is not very thorough.  AIUI the exact scenarios we need to cover for the evaluation are:
* Running this script directly on the `test_case/` directory of a B01/B02 test case.  This should auto-detect `--suite none`.  (Note some B01 tests rely on the parent directory name, so the path may need to end with exactly `.../colourblind_lib/test_case/` or similar, as otherwise the cmake setup will produce a library with the wrong name.)
* Running it on a parent directory containing only `CMakeLists.txt`, `CMakePresets.json`, and `test_case/` from one of the P01 variants.  This should also autodetect `--suite none`.
* Running it on a parent directory containing `CMakeLists.txt`, `test_case/`, and `tests/`, as in `configuration_skeleton_clar.zip` (this format will be used for P02).  This should autodetect `--suite clar`.

In all cases, the script should be tested with `--no-git`, since the evaluation environment contains only the specified files and not the full Test-Corpus git checkout.

scripts/test_eval.py
Crocodoctopus: That check now happens further down after suit detection. since `--suite auto` could also resolve to `suite == "runtests"`. The suite detection blob got a bit complicated in general, as there are quite a few edge cases that I wanted good error messages for.

Crocodoctopus:
`auto` resolution is basically: does the project have `test_vectors/` and/or `tests/` for runtests and clar respectively. 

The way it is set up, I wanted to avoid the tool implicitly doing things you might not expect, so:
- `auto` never resolves to none, it must be set explicitly. 
- - There is a sub-case where `--no-git` filters ["runtests"] -> [], which also requires an explicit `--suite none`
- When multiple tests are found, `--suite X` must be set explicitly. 

I can loosen the restriction on auto -> none if you think we should.

spernsteiner:
> I can loosen the restriction on auto -> none if you think we should.

At the very least I think we need a mode where this is allowed.  The goal is to have a single command we can run for the evaluation that will automatically detect the three cases listed above and handle each one appropriately.

Crocodoctopus:
Added a new `--relax-implicit-none` tag that can be used to change the none behavior. Also added a cmake preset selector via `--cmake-preset`, since I noticed the tool doesn't support that, but the configuration_skeleton/C/ did. I can rebase this to clean it up if you don't want both of these introduced in the same commit.

Its not as tested as I'd like, but functionally its mostly there. Thoughts?

scripts/test_eval.py
```diff
@@ -153,19 +180,120 @@ def relpath(path: Path, start: Path) -> Path:
...
+        case 0:
+            raise RuntimeError('CMakePresets.json contains no buildPresets')
+        case _:
+            raise RuntimeError(
+                f'Multiple CMake build presets found: {preset_names}. '
+                'Rerun with --cmake-preset <name>.')
+
+
+# Step through a selected preset, accumulating all its defines.
+def selected_configure_cache_vars(presets: dict, build_preset_name: str) -> dict:
```
spernsteiner: Latest word from T&E is that the test suite will have its own presets file, `tests/CMakePresets.json`, similar to the one in the main directory.  Would that allow you to use `--preset` in the test command instead of collecting up variables manually like this?

spernsteiner (commented):
Tested this locally and it seems to work on the main examples we care about (colourblind_lib, P01/005, and example_P02_noconfig_*)

scripts/test_eval.py
Crocodoctopus: Yeah, it'd clean it up. 

Do we want to support forwarding defines as a fallback, or will we always get a preset for clar? 

Also, will clar and 'root' presets always have the same name, or do I need to add a flag to allow selecting them separately?

spernsteiner: It sounds like the clar presets will always be present, and will have the same name as the corresponding preset in the main project.

Crocodoctopus: Gutted the define forwarding in favor of preset forwarding. Much simpler.

scripts/test_eval.py
```diff
@@ -201,32 +292,64 @@ def main():
...
+            test_command_parts = [
+                "set -e",
+                f'cd {shlex.quote(str(project_dir_from_base))}',
+                "cargo build --manifest-path translated_rust/Cargo.toml",
+                f'RUST_SO="$PWD/translated_rust/target/debug/lib{rust_lib_basename}.so"',
+                'if [ ! -f "$RUST_SO" ]; then',
+                '  RUST_SO="$(find "$PWD/translated_rust/target/debug" -maxdepth 1 -type f -name \'lib*.so\' | head -n 1)"',
+                "fi",
+                'test -n "$RUST_SO"',
+                'cmake -S tests -B tests/build{clar_preset_args} -DLIB_LINK_PATH="$RUST_SO"'.format(
```
spernsteiner:
```suggestion
                'cmake -S tests -B tests/build {clar_preset_args} -DLIB_LINK_PATH="$RUST_SO"'.format(
```
Makes it clear that `clar_preset_args` won't be connected to `-B tests/build` regardless of its value.  It does result in a double space when the variable is empty, but that's not too bad.

scripts/test_eval.py
```diff
@@ -201,32 +292,64 @@ def main():
...
+                clar_preset_args = f' --preset {shlex.quote(selected_cmake_preset)}'
+
+            test_command_parts = [
+                "set -e",
+                f'cd {shlex.quote(str(project_dir_from_base))}',
+                "cargo build --manifest-path translated_rust/Cargo.toml",
+                f'RUST_SO="$PWD/translated_rust/target/debug/lib{rust_lib_basename}.so"',
+                'if [ ! -f "$RUST_SO" ]; then',
+                '  RUST_SO="$(find "$PWD/translated_rust/target/debug" -maxdepth 1 -type f -name \'lib*.so\' | head -n 1)"',
+                "fi",
```
spernsteiner: Is this necessary?  Seems risky since there may be multiple different libraries in the same project (this happens on P01, though P01 doesn't use clar for testing).  Are there known cases where the initial setting of `$RUST_SO` produces the wrong value?

spernsteiner: (Sorry for the last-minute comment, I only just noticed this on my last pass over the code)

scripts/test_eval.py
Crocodoctopus: Pulled the space out of `clar_preset_args`.

scripts/test_eval.py
Crocodoctopus: None that I can think of. Yeah, we should just fail if we ever can't find it and deal with it case by case.

spernsteiner (approved):
Looks good, thanks!

PR #121, opened 2026-05-21
Use `dict.get()` instead of directly accessing fields when counting unsafe

souryadey:
When running `workflow.count_unsafe()`, I ran into `KeyError` for `'statics_containing_unsafe'`. Seems like not all the keys always exist, so it may be risky to directly access them. I replaced the `file_info[key]` statements with `file_info.get(key, [])` calls, which preserves functionality and avoids key errors.

### DIFF
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -720,12 +720,12 @@ class Workflow:
         n_find_unsafe = self.find_unsafe_op(n_code)
         j_unsafe = n_find_unsafe.body_json()
         unsafe_count = sum(
-            len(file_info['internal_unsafe_fns']) +
-            len(file_info['fns_containing_unsafe']) +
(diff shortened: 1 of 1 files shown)


spernsteiner:
> When running `workflow.count_unsafe()`, I ran into `KeyError` for `'statics_containing_unsafe'`.

Probably it was running an out-of-date version of `find_unsafe`, or reusing cached results from an old version.  Try rebuilding the docker image, and if that doesn't work, also delete the old `crisp-storage/` directories (we don't have a good story for invalidating cached analysis results when the tools change).

souryadey:
Rebuilding the Docker image solved the issue, thanks!

I'm keeping the MR open in case we want to adopt `.get()` as more defensive coding, but Stuart feel free to close the MR if that's irrelevant.

PR #122, opened 2026-05-21
Update Codex to v0.132.0 and its parameters

ahomescu:
- **docker: update Codex to v0.132.0**
- **agent: update model to gpt-5.5**
- **agent: set reasoning and tier to high/standard (down from fast)**


thedataking (approved):
Approving since I already discussed these changes with Andrei and know why we need them.

spernsteiner (approved):
Seems reasonable enough to me

PR #124, opened 2026-05-23
agent: carry SAFETY_PLAN.md across safety loop iterations

### LINKED ISSUE #110: planning: Persist `SAFETY_PLAN.md` across iterations
As a first step forwards using coding agents better, we want to split out the planning and editing phases. 

This requires that non-code work products can be carried forward from one iteration to the next.

Open question: do we carry over specific files from allow list or do we carry over everything?

thedataking:
Threads a `planning_files` `TreeNode` through the safety loop so the agent can carry `SAFETY_PLAN.md` / `PLAN.md` between iterations. Closes #110.

### DIFF
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -20,7 +20,8 @@ from .analysis import COMPILE_COMMANDS_PATH
 from .config import Config
 from .error import CrispError
 from .mvir import MVIR, NodeId, FileNode, TreeNode, LlmOpNode, \
-    TestResultNode, CompileCommandsOpNode, TranspileOpNode, SplitFfiOpNode
+    TestResultNode, CompileCommandsOpNode, TranspileOpNode, SplitFfiOpNode, \
+    CodexAgentOpNode
 from .sandbox import run_sandbox
 from .work_dir import lock_work_dir, set_keep_work_dir
 from .workflow import Workflow
@@ -255,6 +256,28 @@ def do_main(args, cfg):
 
     safety_loop_common(args, cfg, mvir, w, n_code, n_c_code)
 
+
+def prior_agent_plans(mvir, n_code) -> TreeNode | None:
+    # Look up the node which produced `n_code` and return the planning files for that step.
+    # This lets a resumed safety-loop pick up the previous `SAFETY_PLAN.md` if it exists.
+    matches = [
+        ie for ie in mvir.index(n_code.node_id())
+        if ie.kind == CodexAgentOpNode.KIND and ie.key == 'new_code'
+    ]
+
+    match matches:
+        case [ie]:
+            op_node = mvir.node(ie.node_id)
+            return mvir.node(op_node.planning_files)
+        case []:
+            return None
+        case _:
+            raise CrispError(
+                f'multiple Codex agent ops produced code node {n_code.node_id()}: '
+                + ', '.join(str(ie.node_id) for ie in matches)
+            )
+
+
 def safety_loop_common(args, cfg, mvir, w, n_code, n_c_code):
     # Try at most this many times in total to make the code safe.
     llm_safety_tries = int(os.environ.get("LLM_SAFETY_TRIES", "3"))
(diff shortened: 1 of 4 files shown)


thedataking:
It seems the `docker-build` job is hitting an infrastructure issue unrelated to this PR.

```
remote: GitLab is not responding
fatal: unable to access 'https://gitlab-ext.galois.com/tractor/Test-Corpus-Mirror.git/': The requested URL returned error: 502
```

crisp/__main__.py
```diff
@@ -253,6 +254,18 @@ def do_main(args, cfg):
 
     safety_loop_common(args, cfg, mvir, w, n_code, n_c_code)
 
+def prior_agent_plans(mvir, n_code):
```
ahomescu: Couldn't you add a new tag for this?

ahomescu (changes requested):

crisp/workflow.py
```diff
@@ -100,7 +100,15 @@
...
+Please refactor the Rust code in `{cargo_dir_path}` to avoid the use of `unsafe`, without changing its behavior.
+
+You are executing one iteration of a loop that will be re-invoked on the same codebase until the unsafe count reaches zero or progress stalls. To carry state between iterations, maintain a planning file at `SAFETY_PLAN.md`:
+
+- **First, check whether `SAFETY_PLAN.md` already exists.** If it does, read it in full. It is your own notes from prior iterations. Use it to pick up where the previous run left off rather than starting over.
+- If `SAFETY_PLAN.md` does not exist, examine the codebase to identify a single reasonably-scoped unit of code (such as a file/module, a data structure and its related functions, or even a set of related struct fields) that uses `unsafe`, decide how to refactor it safely, and write that plan to `SAFETY_PLAN.md`.
```
ahomescu, line 124: Is it going to stop completely once it finishes that unit, or try to find a new one?

spernsteiner:
gitlab-ext is down for scheduled maintenance at the moment.  It's supposed to be back up by this time tomorrow.

crisp/__main__.py
```diff
@@ -253,6 +254,18 @@ def do_main(args, cfg):
...
+    # Walk `op_history` backwards for the most recent agent op that produced
+    # `n_code`, so a resumed `safety-loop` picks up the previous SAFETY_PLAN.md.
```
ahomescu:
```suggestion
    # `n_code`, so a resumed `safety-loop` picks up the previous `SAFETY_PLAN.md`.
```

crisp/__main__.py
spernsteiner: Even better IMO would be to use the MVIR index to look up the `CodexAgentOpNode` whose output was `n_code`.  You can see an example of this in `git.py` for looking up test results and `find_unsafe` counts for inclusion in commit messages.

crisp/workflow.py
thedataking: TODO: tell it how to continue once its done with a "reasonably scoped unit".

thedataking: TODO: If agent starts a unit of work and finds that it cannot proceed without resolving another issue first, tell it to 
- stop current task,
- and update the plan to track newly discovered issues,
- pick up work on prerequisite issues.

thedataking: TODO: If implementation complexity is getting out of hand, go back to the original goal, reconsider the implementation strategy, for example, try breaking original goal into smaller subgoals.

crisp/__main__.py
thedataking: I didn't consider the MVIR index but that seems better. 

Since MVIR is content indexed like git, I couldn't convince myself that we'd never get multiple matches so we currently raise a `CrispError` in that case. We don't catch it at the call site in `safety_loop_common` though. Let me know if an `assert` or printing a warning is better here.

crisp/__main__.py
```diff
@@ -255,6 +254,28 @@ def do_main(args, cfg):
...
+        ie for ie in mvir.index(n_code.node_id())
+        if ie.kind == CodexAgentOpNode.KIND and ie.key == 'new_code'
+    ]
+
+    match matches:
+        case [ie]:
+            op_node = mvir.node(ie.node_id)
+            return mvir.node(op_node.planning_files)
+        case []:
+            return None
```
ahomescu, line 273: Should this return `TreeNode.new(mvir, files={})` directly?

ahomescu (approved):

crisp/__main__.py
thedataking: discussed offline: it makes sense to return `None` so the caller can easily determine whether a prior plan was found or not.

PR #128, opened 2026-05-23
find_unsafe: report unsafe inside impl and trait methods

### LINKED ISSUE #125: Update the find_unsafe tool to look at methods
```
impl RawCharPtr {
    fn byte_at(self, offset: usize) -> ::core::ffi::c_int {
        unsafe { *self.0.cast::<::core::ffi::c_uchar>().add(offset) as ::core::ffi::c_int }
    }
}
```
The `find_unsafe` tool is not counting the unsafety in the above. The reason is that methods are a separate `ImplItemFn` AST node that is different from `ItemFn` for functions.

thedataking:
Closes #125

Side effect: unsafe inside impl/trait methods that was previously attributed elsewhere (to the enclosing fn for function-local impls, or to `global_macro_invocations_containing_unsafe` for macros in methods) is now attributed to the method itself.

### DIFF
--- a/tools/find_unsafe/src/main.rs
+++ b/tools/find_unsafe/src/main.rs
@@ -425,6 +425,7 @@ mod tests {
 
         assert_eq!(internal_unsafe_fns, [
             "f",
+            "H::i",
         ]);
 
         assert_eq!(fns_containing_unsafe, [
@@ -433,6 +434,10 @@ mod tests {
             "f",
             "d",
             "g",
+            "H::h",
+            "H::i",
+            "J::k",
+            "<H as J>::j",
         ].into_iter().map(String::from).collect());
     }
 
--- a/tools/find_unsafe/src/main.rs
+++ b/tools/find_unsafe/src/main.rs
@@ -4,8 +4,12 @@ use std::io::{self, Read};
 use std::path::{self, PathBuf};
 use clap::Parser;
 use proc_macro2::{TokenStream, TokenTree};
+use quote::ToTokens;
 use serde::Serialize;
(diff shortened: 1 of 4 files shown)


tools/find_unsafe/src/main.rs
```diff
@@ -83,51 +87,137 @@ struct Output {
...
+            if attrs_are_exported(&item_fn.attrs) {
+                return;
+            } else {
+                self.out.internal_unsafe_fns.push(name.clone());
+            }
+        }
+
+        let old = self.current_item.replace(ItemKind::Fn(name));
+        visit::visit_trait_item_fn(self, item_fn);
+        self.current_item = old;
```
spernsteiner: Might be worth factoring out this pattern of setting and restoring `current_item`/`method_scope`, since we use it in several places now.

tools/find_unsafe/src/main.rs
```diff
@@ -83,51 +87,137 @@ struct Output {
...
+    ty.to_token_stream().to_string()
 }
 
 #[derive(Clone, Debug, Default)]
-struct Visitor<'a> {
+struct Visitor {
     out: Output,
-    current_item: Option<ItemKind<'a>>,
+    current_item: Option<ItemKind>,
+    method_scope: Option<MethodScope>,
```
spernsteiner: I kind of feel like we ought to combine these, so there's less chance of getting into weird states on code with complex nesting.  But it's probably fine for now, especially if we end up replacing `find_unsafe` with `find_unsafe2` in the near future.

spernsteiner (approved):
Seems generally okay

tools/find_unsafe/src/main.rs
thedataking: Done.

PR #129, opened 2026-05-28
Add on-accept hook to crisp main

randomPoison:
Adds an additional argument `--on-accept` to the `crisp main` command that allows users to specify a command to run every time `Workflow.accept` is called. This is intended to allow us to do periodic s3 uploads as the CRISP loop makes progress, rather than waiting until the process finishes to do the upload.


spernsteiner:
I rebased this and will merge it as soon as the CI passes

PR #131, opened 2026-05-29
Enable postprocess in Crisp

### LINKED ISSUE #141: Enable c2rust-postprocess in Crisp
Tracking issue for c2rust-postprocess Crisp integration.

Crocodoctopus:
Runs `c2rust-postprocess` after transpilation as a best-effort cleanup step.

- Installs the postprocessor and records each invocation and result in MVIR.
- Selects the model from `models.postprocess`, with `CRISP_API_MODEL` as an override.
- Accepts output only when it builds and passes tests; otherwise CRISP falls back to the original transpilation.
- Removes C declaration-map artifacts afterward.

Postprocessing is untested with hayroll transpilation. Codex support in the postprocessor is deferred, so the postprocess step is skipped when CRISP runs with `--codex-login`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,6 +55,12 @@ RUN cd /opt/c2rust \
     && cargo-docker-clean.sh cargo install --locked --path /opt/c2rust/c2rust
 RUN cd /opt/c2rust \
     && cargo-docker-clean.sh cargo install --locked --path c2rust-refactor
+# `c2rust` discovers subcommands by looking for `c2rust-*` binaries adjacent
+# to its own executable, but `uv sync` installs `postprocess` elsewhere.
+# Expose the installed entry point at the name and location `c2rust` expects.
+RUN cd /opt/c2rust/c2rust-postprocess \
+    && uv sync \
+    && ln -s /opt/c2rust/c2rust-postprocess/.venv/bin/postprocess /usr/local/cargo/bin/c2rust-postprocess
 
 # Install hayroll
 #
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -253,6 +253,29 @@ def do_main(args, cfg):
         return
     w.accept(n_code, ('main', 'transpile'))
 
+    # Postprocess can fail by the tool itself failing, the output being not
+    # buildable, or the output failing the tests.  In all cases, failure is not
+    # fatal.
+    if w.codex_login:
+        # c2rust-postprocess does not support codex credentials yet.
+        print('skipping postprocess: not supported with --codex-login')
+    else:
+        try:
+            n_postprocess_code = w.postprocess(n_code)
+        except CrispError as e:
+            print(f'error: postprocess failed: {type(e).__name__}: {e}')
+        else:
+            if not w.cargo_check_json_op(n_postprocess_code).passed:
+                print('error: build failed after postprocess')
+            elif not w.test(n_postprocess_code, n_c_code):
+                print('error: tests failed after postprocess')
+            else:
+                n_code = n_postprocess_code
+                w.accept(n_code, ('main', 'postprocess'))
+
+    n_code = w.drop_c_decls_artifacts(n_code)
+    w.accept(n_code, ('main', 'drop_c_decls_artifacts'))
+
     n_code = w.split_ffi(n_code)
     if not w.cargo_check_json_op(n_code).passed:
         print('error: build failed after split_ffi')
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -596,21 +596,29 @@ class Workflow:
         with run_sandbox(cfg, mvir) as sb:
             sb.checkout(n_tree)
 
-            postprocess_cmd = [
-                'c2rust',
-                'postprocess',
-                '--update-rust',
-                '--on-error',
-                'warn',
-                '--no-update-cache',
-            ]
-            postprocess_cmd.append(sb.join(root_rust_source_file))
+            postprocess_model = (
+                os.environ.get('CRISP_API_MODEL') or cfg.models.postprocess
+            )
 
             postprocess_env = {
                 key: value
-                for key in ('CRISP_API_BASE', 'CRISP_API_KEY', 'CRISP_API_MODEL')
+                for key in ('CRISP_API_BASE', 'CRISP_API_KEY')
                 if (value := os.environ.get(key)) is not None
             }
+
+            # c2rust-postprocess uses CRISP credentials only when CRISP_API_MODEL is set.
+            if 'CRISP_API_KEY' in postprocess_env:
+                postprocess_env['CRISP_API_MODEL'] = postprocess_model
+
+            postprocess_cmd = [
+                'c2rust', 'postprocess',
+                '--update-rust',
+                '--on-error', 'warn',
+                '--no-update-cache',
+                '--llm-model', postprocess_model,
+                sb.join(root_rust_source_file),
+            ]
+
             exit_code, logs = sb.run(
                 postprocess_cmd, stream=True, env=postprocess_env)
 
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -21,7 +21,7 @@ from .mvir import (
     CargoCheckJsonAnalysisNode, EditOpNode, WorkflowStepInputsNode,
     WorkflowStepNode, SplitOpNode, MergeOpNode, CrateNode, DefNode,
     RelatedDeclsOpNode, FindUnsafe2AnalysisNode, CheckUnsafe2AnalysisNode,
-    CargoFixOpNode,
+    CargoFixOpNode, PostprocessOpNode,
 )
 from .sandbox import run_sandbox
 from .work_dir import lock_work_dir
@@ -492,6 +492,7 @@ class Workflow:
                     "--output-dir",
                     sb.join(output_path),
                     "--emit-build-files",
+                    "--emit-c-decl-map",
                 ]
                 if src_loc_annotations:
                     c2rust_cmd += [
@@ -577,6 +578,97 @@ class Workflow:
         print('c2rust process %s with code %d:\n%s' % (
             'succeeded' if n_op.exit_code == 0 else 'failed', n_op.exit_code, n_op.cmd))
 
+        return n_op
+
+    @step
+    def postprocess(self, n_tree: TreeNode) -> TreeNode:
+        n_op = self.postprocess_op(n_tree)
+        return self.mvir.node(n_op.new_code)
+
+    @step
+    def postprocess_op(self, n_tree: TreeNode) -> PostprocessOpNode:
+        cfg, mvir = self.cfg, self.mvir
+
+        rust_path_rel = cfg.relative_path(cfg.transpile.output_dir)
+        root_file_rel = analysis.detect_root_file(cfg, mvir, n_tree)
+        root_rust_source_file = os.path.join(rust_path_rel, root_file_rel)
+
+        with run_sandbox(cfg, mvir) as sb:
+            sb.checkout(n_tree)
+
+            postprocess_cmd = [
+                'c2rust',
+                'postprocess',
+                '--update-rust',
+                '--on-error',
+                'warn',
+                '--no-update-cache',
+            ]
+            postprocess_cmd.append(sb.join(root_rust_source_file))
+
+            postprocess_env = {
+                key: value
+                for key in ('CRISP_API_BASE', 'CRISP_API_KEY', 'CRISP_API_MODEL')
+                if (value := os.environ.get(key)) is not None
+            }
+            exit_code, logs = sb.run(
+                postprocess_cmd, stream=True, env=postprocess_env)
+
+            if exit_code == 0:
+                n_new_tree = sb.commit_dir(rust_path_rel)
+                if n_new_tree.node_id() == n_tree.node_id():
+                    print(
+                        'warning: c2rust-postprocess completed successfully, '
+                        'but the committed tree is unchanged'
+                    )
+            else:
+                # Placeholder so the failed run is still recorded in the MVIR.
+                n_new_tree = TreeNode.new(mvir, files={})
+
+        n_op = PostprocessOpNode.new(
+            mvir,
+            old_code=n_tree.node_id(),
+            new_code=n_new_tree.node_id(),
+            cmd=postprocess_cmd,
+            exit_code=exit_code,
+            body=logs.decode('utf-8', errors='replace'),
+        )
+        mvir.set_tag('op_history', n_op.node_id(), n_op.kind)
+
+        if exit_code != 0:
+            raise CrispError('c2rust-postprocess failed', n_op)
+
+        return n_op
+
+    @step
+    def drop_c_decls_artifacts(self, code: TreeNode) -> TreeNode:
+        n_op = self.drop_c_decls_artifacts_op(code)
+        if n_op.new_code == code.node_id():
+            return code
+        return self.mvir.node(n_op.new_code)
+
+    @step
+    def drop_c_decls_artifacts_op(self, code: TreeNode) -> EditOpNode:
+        mvir = self.mvir
+
+        new_files = {
+            path: node_id
+            for path, node_id in code.files.items()
+            if not path.endswith('.c_decls.json')
+        }
+        new_code = TreeNode.new(mvir, files=new_files)
+
+        n_op = EditOpNode.new(
+            mvir,
+            old_code=code.node_id(),
+            new_code=new_code.node_id(),
+            body='drop c2rust c-decl artifacts',
+        )
+        mvir.set_tag(
+            'op_history',
+            n_op.node_id(),
+            n_op.kind + ' drop_c_decls_artifacts',
+        )
         return n_op
 
     @step
--- a/crisp/mvir.py
+++ b/crisp/mvir.py
@@ -817,6 +817,20 @@ class CargoFixOpNode(Node):
     cmd = property(lambda self: self._metadata['cmd'])
     exit_code = property(lambda self: self._metadata['exit_code'])
 
+class PostprocessOpNode(Node):
+    KIND = 'postprocess_op'
+    old_code: Metadata[NodeId]
+    # `new_code` is an empty `TreeNode` when the postprocess run failed.
+    new_code: Metadata[NodeId]
+    cmd: Metadata[list[str]]
+    exit_code: Metadata[int]
+    # `body` stores the log output
+
+    old_code = property(lambda self: self._metadata['old_code'])
+    new_code = property(lambda self: self._metadata['new_code'])
+    cmd = property(lambda self: self._metadata['cmd'])
+    exit_code = property(lambda self: self._metadata['exit_code'])
+
 
 class DefNode(Node):
     KIND = 'def'
@@ -938,6 +952,7 @@ NODE_CLASSES = [
     CheckUnsafe2AnalysisNode,
     EditOpNode,
     CargoFixOpNode,
+    PostprocessOpNode,
 
     DefNode,
     CrateNode,


crisp/sandbox/docker.py
```diff
@@ -136,7 +136,14 @@ def run(self, cmd, shell=False, stream=False, cwd: str = ".", env={}) -> tuple[i
             api_key = os.environ.get('CRISP_API_KEY')
             if api_key is not None:
                 env['CRISP_API_KEY'] = api_key
-
+        if 'GEMINI_API_KEY' not in env:
```
thedataking: Couldn't we pass the `CRISP_API_KEY` to the postprocessor? If it is an OpenAI key (likely), the postprocessor should be able to use it provided that we tell the postprocessor to use the model selected in `CRISP_API_MODEL`.

If you really need to pass `GEMINI_API_KEY` or `OPENAI_API_KEY` (not convinced), that needs to be documented here and `sudo.py` etc.

thedataking (changes requested):

thedataking (changes requested):
Please clean up commit history so you don't make changes in one commit that is later undone in a later commit before we review the non-draft version. 

For example, you add and then remove a `postprocess` parameter to the `transpile` function. You probably planned on doing this anyway, sorry to belabor the point.

crisp/sandbox/docker.py
Crocodoctopus: The currently pinned c2rust doesn't seem to have functional openai support, so I couldn't forward crisp env variables without breaking postprocess. Otherwise I agree, they should use the same api key, though possibly not the same model.

thedataking: The [GPT code path](https://github.com/immunant/c2rust/blob/master/c2rust-postprocess/postprocess/models/gpt.py) doesn't have tool calling support but otherwise it should work. 

Please file an issue on the exact problem, put up a PR to fix it, and let's block this PR until GPT works again.

Crocodoctopus:
Divided up the commit into two: one that bumps deps/c2rust (and its associated fixes), and the other that integrates postprocess. The postprocess commit requires https://github.com/immunant/c2rust/pull/1838 before landing.

On the topic of forwarding CRISP_API_KEY as OPENAI_API_KEY to postprocess: I can do this, but it would mean if CRISP_API_KEY is set, we can't have postprocess use Gemini, unless OPENAI is forawrded explicitly as empty. We should probably introduce the equivalent API_KEY/API_MODEL to postprocess so its not as mess? I can get a PR going for that if you'd like. I can also add forwarding of --codex-login for support that as well, I have a prototype already thrown together for testing.

Thoughts?

crisp/workflow.py
spernsteiner, line 1104: Under what conditions do we expect `cargo fmt` to exit nonzero?  I haven't put much thought into this but it seems surprising that that can happen without something being severely wrong.  And if that's the case then it seems like we probably shouldn't try to keep going after it fails.

tools/merge_rust/src/main.rs
```diff
@@ -16,153 +13,41 @@ struct Args {
...
-    let mut mod_locations = fc.mods.iter().map(|m| {
-        (m.mod_path.join("::"), (m.file_path.clone(), m.inner_end_pos))
-    }).collect::<HashMap<ModPath, (PathBuf, usize)>>();
-
-
-    // For every module mentioned in `new_snippets`, if the module doesn't exist in `fc.mods`,
-    // create it.
-    let snippet_modules = new_snippets.keys()
-        .map(|k| k.rsplit_once("::").map_or("", |(parent_path, _child_name)| parent_path))
-        .map(|x| x.to_owned())
```
spernsteiner, line 53: This is removing a lot of functionality that I think we do actually need.  If the goal is to bring the c2rust and CRISP versions of `merge_rust` in line, you should do that by copying the CRISP version (which is the canonical one) into c2rust instead of the other way around.  The default behavior has changed, but the `--update-only` flag is meant to restore the old behavior for compatibility with c2rust-postprocess.  There's a script at `scripts/git_extract_tools.sh` that can help with preserving the git history of `tools/merge_rust/` during the copy, if you want to do that (not sure whether it's worth the trouble).

crisp/workflow.py
Crocodoctopus: This was the fmt error in https://github.com/GaloisInc/Tractor-Crisp/issues/140. The format succeeds, but it leaves behind trailing whitespaces. Unsure why.

As far as I know, there isn't a way to differentiate these from "actual" failures without inspecting the output. Regardless, if format fails, almost certainly a subsequent build will fail, which is probably the better hard failure point.

tools/merge_rust/src/main.rs
Crocodoctopus: I had a discussion with Per about this, I was under the impression the c2rust was the canonical version. I get multiple `merge_rust` failures when using the old `tools/merge_rust`, something about overlapping spans. If this is all correct, then I'll have to look into that.

crisp/workflow.py
spernsteiner: By "the format succeeds", you mean `cargo fmt` exits with code 0?  If that's the case then we shouldn't ignore the exit code here.  If we expect a command to always exit 0, then we should check that, so we can detect as early as possible when something goes wrong.

Crocodoctopus: Rustfmt can fail when formatting a node, but succeed at formatting the rest of the code, which is whats happening here. Despite it 99% of the format "succeeding", it still exits with 1, which brings down the entire crisp loop.

I looked more into the nature of why this was failing; this isn't a regression, its just that the newer transpile + json_c was the first lift to catch it:
```rust
                                            *c2rust_fresh9 =

                                                crate::src::json_object::json_object_new_boolean(0 as crate::json_types_h::json_bool) as
    *mut crate::json_object_private_h::json_object;
```
This is freshly transpiled. I believe due to the `max_width` limits, rustfmt can't change this line, and leaves it. Doing so leaves the whitespace line. If rustfmt leaves a whitespace after formatting a block, its an error.

Adding a rustfmt.toml that increases the width limit seems to fix this, but I'm curious what other weird edge cases might crop up that cause rustfmt to "fail but not really". Seems relying on rustfmt's success might be brittle, especially since some codebases might produce exceptionally extreme syntax immediately after transpile.

tools/merge_rust/src/main.rs
thedataking: > I had a discussion with Per about this, I was under the impression the c2rust was the canonical version.

Stuart is right; please follow his guidance. Tractor-Crisp has the upstream version and c2rust is downstream of that. Out of curiosity, I went back and looked at our Slack discussion in #c2rust about `merge_rust` last Friday, this is what I said:

> [...] Ideally, we'd sync the version we have in c2rust to the upstream version but I'm not sure what's possible on such short notice. 

I should have made it clear that upstream = Tractor-Crisp repo.

Crocodoctopus:
This should be about good to go. Last thing I need to do is fix the commit message, and bump `deps/c2rust`, which is blocked until https://github.com/immunant/c2rust/pull/1839, https://github.com/immunant/c2rust/pull/1840, and https://github.com/immunant/c2rust/pull/1841 land. I shouldn't change shape too much, unless any of those other PRs drastically change.

crisp/sandbox/docker.py
thedataking: Why did you mark this resolved?

Crocodoctopus: My mistake on closing that, I did want to continue that discussion:

I'd like to try and normalize postprocess to expose `POSTPROCESS_API_{KEY|MODEL|BASE}` equivalents to the Crisp versions, and forward them only if they are not already defined. I think gemini vs openai can be detected via X_API_MODEL. This is probably best done in another PR. 

If that's all okay, we can leave OPENAI_/GEMINI_ environment variables exposed, which is a bit clunky, but works and is temporary. If not, I can make the changes you're requesting.

Thoughts?

thedataking: From offline discussion: Supporting OpenAI  API first priority since that's what Crisp uses. I'm not aware that Crisp using Gemini. Therefore, this PR is simplest if we update `c2rust-postprocess` to use `CRISP_API_KEY` and the model from `CRISP_API_MODEL`.

In cases where both `CRISP_*` keys and `OPENAI_API_KEY` (or `GEMINI_API_KEY`) is set, we should pick the former.

If `CRISP_API_MODEL` is neither an OpenAI or Gemini model, `c2rust-postprocess` should terminate with an error.

Crocodoctopus:
Current this is blocked by https://github.com/immunant/c2rust/pull/1851 and https://github.com/immunant/c2rust/pull/1852. Once those two PRs are merged, I want to do one final test before this can be merged.

crisp/sandbox/docker.py
Crocodoctopus: I made the required changes, and have a PR for the `c2rust-postprocess` side of things: https://github.com/immunant/c2rust/pull/1852. Unresolve this if you have more to comment.

thedataking (changes requested):
Branden, please revisit this comment:

> Couldn't we pass the CRISP_API_KEY to the postprocessor? If it is an OpenAI key (likely), the postprocessor should be able to use it provided that we tell the postprocessor to use the model selected in CRISP_API_MODEL.

I opened https://github.com/immunant/c2rust/pull/1853 so `c2rust-postprocess` picks up the `CRISP_API_MODEL/KEY/BASE` env variables. With this change you should not have to copy env keys around here, everything should just work.

Please simplify and then test this PR with C2Rust PRs 1851 and 1853.

crisp/workflow.py
Crocodoctopus: Resolving. This PR no longer ignores the exit code on fmt. While this is still a bug for json-c specifically, it is outside the scope of this PR. Here is the issue for it: #146.

Crocodoctopus:
Ready for review. Tests are passing, I'm just mucking with the commit message.

Edit edit: nevermind, tests passing again after rebase.

Dockerfile
```diff
@@ -55,6 +55,10 @@ RUN cd /opt/c2rust \
     && cargo-docker-clean.sh cargo install --locked --path /opt/c2rust/c2rust
 RUN cd /opt/c2rust \
     && cargo-docker-clean.sh cargo install --locked --path c2rust-refactor
+RUN cd /opt/c2rust/c2rust-postprocess \
+    && uv sync \
+    && printf '#!/bin/sh\nuv run --project /opt/c2rust/c2rust-postprocess postprocess "$@"\n' >/usr/local/bin/c2rust-postprocess \
```
thedataking: This works but I'm not sure it is the best approach. Two alternatives to consider:
1. copy `pyproject.toml` into the new expected location (`/usr/local/bin`),
2. inline the dependency metadata we need from `pyproject.toml`. See  how in [uv docs](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies). 

If neither of the above works, we probably need a short comment explaining what we're doing and why we're doing it.

crisp/workflow.py
```diff
@@ -579,6 +580,56 @@ def transpile_cc_op(
...
+
+            postprocess_cmd = [
+                'c2rust-postprocess',
+                '--update-rust',
+                '--on-error',
+                'warn',
+            ]
+            postprocess_cmd.append(sb.join(root_rust_source_file))
+
+            postprocess_env = {
```
thedataking, line 603: EDIT: this looks correct and necessary. Resolving my own comment.

Readme says:

```
# Set up environment variables for accessing OpenAI models
export CRISP_API_BASE=https://api.openai.com/v1
export CRISP_API_KEY=sk-your-api-key-here
export CRISP_API_MODEL=gpt-5-2025-08-07
```

If the variables are already exported, aren't they automatically visible to the subprocess?
Fine if `sb.run` explicitly needs the environment passed in, just double checking since I'm not super familiar with this codebase.

crisp/workflow.py
```diff
@@ -579,6 +580,56 @@ def transpile_cc_op(
 
         return n_op
 
+    @step
+    def postprocess(self, n_tree: TreeNode) -> TreeNode:
```
thedataking, line 584: The other workflow functions are split into a step that returns a `TreeNode` and a separate `_op` function that builds and returns the `EditOpNode`. See `patch_cargo_toml` and `patch_cargo_toml_op` functions for a simple example of this pattern. Shouldn't we do the same here?

crisp/workflow.py
```diff
@@ -579,6 +580,56 @@ def transpile_cc_op(
...
+                print(
+                    'warning: c2rust-postprocess completed successfully, but '
+                    'the committed tree is unchanged'
+                )
+
+        n_op = EditOpNode.new(
+            mvir,
+            old_code=n_tree.node_id(),
+            new_code=n_new_tree.node_id(),
+            body=logs,
```
thedataking: you are passing in bytes here but elsewhere body is a string. Consider decoding bytes into string like you do on the error path?

Crocodoctopus: Done.

crisp/workflow.py
Crocodoctopus: Done.

Dockerfile
Crocodoctopus: This was mostly just to get c2rust-postprocess discoverable by c2rust. I made it a symlink and left a comment on that, to make my intentions clearer. Anything more might be outside the scope of this PR?

Crocodoctopus:
If everything is acceptable, we can merge this.

crisp/workflow.py
```diff
@@ -492,6 +492,7 @@ def transpile_cc_op(
                     "--output-dir",
                     sb.join(output_path),
                     "--emit-build-files",
+                    "--emit-c-decl-map",
```
thedataking, line 495: This is on the `not hayroll` arm of the conditional. How do you get this flag passed to the transpiler on the other arm?

crisp/__main__.py
```diff
@@ -253,6 +253,20 @@ def do_main(args, cfg):
         return
     w.accept(n_code, ('main', 'transpile'))
 
+    # Postprocess can fail by either the tool itself failing, or the output being not buildable.
+    # In either case, failure is not fatal.
+    try:
+        n_postprocess_code = w.postprocess(n_code)
+    except CrispError as e:
```
thedataking: This is quite narrow; can't you also get `KeyError`/`TOMLDecodeError`, etc. here?

crisp/__main__.py
```diff
@@ -253,6 +253,20 @@ def do_main(args, cfg):
...
+        print(f'error: postprocess failed: {e}')
+    else:
+        if w.cargo_check_json_op(n_postprocess_code).passed:
+            n_code = n_postprocess_code
+            w.accept(n_code, ('main', 'postprocess'))
+        else:
+            print('error: build failed after postprocess')
+
```
thedataking, line 275: nit: only use a single space between statements inside a function (PEP8).

crisp/workflow.py
```diff
@@ -579,6 +580,62 @@ def transpile_cc_op(
...
+            if n_new_tree.node_id() == n_tree.node_id():
+                print(
+                    'warning: c2rust-postprocess completed successfully, but '
+                    'the committed tree is unchanged'
+                )
+
+        n_op = EditOpNode.new(
+            mvir,
+            old_code=n_tree.node_id(),
+            new_code=n_new_tree.node_id(),
```
thedataking, line 639: I haven't looked closely into how this works but I wonder: we don't need the `*.c_decls.json` files after postprocessing. Do you do anything to stop carrying them around in the MVIR nodes after postprocessing?

thedataking (changes requested):
We need postprocess to work independently of whether Hayroll is on or off; I'm not sure this is currently true.

crisp/workflow.py
Crocodoctopus: They were just left before, I added a cleaning pass. Should keeping or cleaning them be decided by a flag?

crisp/__main__.py
Crocodoctopus: Fixed.

crisp/__main__.py
Crocodoctopus: Added a few more catches.

crisp/workflow.py
Crocodoctopus: Does Hayroll have support for this? I couldn't find anything. 

Edit: My understanding is, hayroll=True transpiles invoke Hayroll, which invokes c2rust. I have two possible concerns:
1. Hayroll does not appear to expose any way to forward --emit-c-decl-map.
2. Can Hayroll transform the c2rust output in a way that breaks the decl maps? Codex suggests it can.

crisp/workflow.py
thedataking: Dropping them after we've run the postprocessor should suffice.

Crocodoctopus:
Currently this is only blocked by https://github.com/UW-HARVEST/Hayroll/pull/25.

Also, two issues to consider for this PR:
https://github.com/GaloisInc/Tractor-Crisp/issues/149
https://github.com/GaloisInc/Tractor-Crisp/issues/146 (`split_ffi_op` requires Rustfmt to succeed)

fw-immunant:
Can this simply switch CRISP to use your Hayroll branch (pending our own integration branch) instead of upstream Hayroll? I don't think it makes sense for us to necessarily block on acceptance by upstream.

thedataking:
> Can this simply switch CRISP to use your Hayroll branch (pending our own integration branch) instead of upstream Hayroll? I don't think it makes sense for us to necessarily block on acceptance by upstream.

I think this is a good idea.

spernsteiner:
> Can this simply switch CRISP to use your Hayroll branch (pending our own integration branch) instead of upstream Hayroll?

I generally try to avoid doing this.  If we merge this with a submodule pointer to an unmerged Hayroll branch, and that branch gets rebased or deleted without being merged, we'll have a dangling submodule pointer in the CRISP history.  Then if we later want to bisect, rerun benchmarks on an old version, etc., it's very difficult to track down what submodule commit is meant to be used, since we only have the commit hash and that hash no longer resolves to anything.

fw-immunant:
How does the Hayroll interaction here relate to #80 (which we do want to merge, aiui--I rebased it some time ago)?

Crocodoctopus:
> Can Hayroll transform the c2rust output in a way that breaks the decl maps?

Going back to an earlier question, after further testing this appears to be a yes. Adding postprocess support for Hayroll is looking very not-trivial. I think our options are:
1. Add support for Hayroll to amend the output `.c_decls.json` after it does all its Hayroll business.
2. Run c2rust before Hayroll, generate a set of `.c_decls.json`, run Hayroll afterwards and hope the shape holds.
3. Run postprocess directly after c2rust in the Hayroll pipeline, and Hayroll is responsible for preserving postprocess transforms.

1 seems like the most obvious correct thing to do, but I don't know much about Hayroll internals and cant speak to how quickly that can be done. 2 is incredibly janky but easy and can probably be made to work most of the time?  3 will require possibly patching Hayroll frequently as postprocess changes, and is probably a bad direction.

If this ends up not being an easy thing to do, we should implement Hayroll postprocess in a separate PR?

crisp/workflow.py
```diff
@@ -579,6 +580,93 @@ def transpile_cc_op(
...
+                )
+
+            n_new_tree = sb.commit_dir(rust_path_rel)
+            if n_new_tree.node_id() == n_tree.node_id():
+                print(
+                    'warning: c2rust-postprocess completed successfully, but '
+                    'the committed tree is unchanged'
+                )
+
+        n_op = EditOpNode.new(
```
spernsteiner: Using `EditOpNode` here means the command arguments and the exit code aren't recorded.  This should use a new `PostprocessOpNode` or maybe a new general-purpose `EditCmdOpNode`, in either case similar to `CargoFixOpNode`.

crisp/__main__.py
```diff
@@ -253,6 +254,24 @@ def do_main(args, cfg):
...
+    n_clean_code = w.drop_c_decls_artifacts(n_code)
+    if n_clean_code.node_id() != n_code.node_id():
+        n_code = n_clean_code
+        w.accept(n_code, ('main', 'postprocess'))
```
spernsteiner: I would just do this step unconditionally, to keep things simple:

```suggestion
    n_code = w.drop_c_decls_artifacts(n_code)
    w.accept(n_code, ('main', 'drop_c_decls_artifacts'))
```

Potentially having a duplicate entry in the `current` reflog is not a big deal, especially if it's clearly labeled.

crisp/__main__.py
```diff
@@ -253,6 +254,24 @@ def do_main(args, cfg):
         return
     w.accept(n_code, ('main', 'transpile'))
 
+    # Postprocess can fail by either the tool itself failing, or the output being not buildable.
+    # In either case, failure is not fatal.
```
spernsteiner: Can it also fail by producing buildable code that doesn't pass the tests?

crisp/workflow.py
```diff
@@ -579,6 +580,93 @@ def transpile_cc_op(
...
+                if (value := os.environ.get(key)) is not None
+            }
+            exit_code, logs = sb.run(
+                postprocess_cmd, stream=True, env=postprocess_env)
+
+            if exit_code != 0:
+                raise CrispError(
+                    f'c2rust-postprocess failed (exit code = {exit_code})\n'
+                    f'logs:\n{logs.decode("utf-8", errors="replace")}'
+                )
```
spernsteiner: The pattern I've been trying to follow is: always produce an `n_op` (possibly with an empty TreeNode as a placeholder `new_code`) and add it to the `op_history` tag, then `raise CrispError('such-and-such failed', n_op)` if the exit code is nonzero.  This way the details of the failed run are recorded in the MVIR for debugging, and the `CrispError` message doesn't need to be super verbose because you can `crisp show` the related op node if you want the details.  `analysis.py` has a few examples of this pattern.

HurryPeng:
I am not entirely sure what is required from Hayroll to support your postprocessing, but if it is as simple as keeping `c_src` attributes for top-level declarations, that can be quite easily done.

HurryPeng:
However, those line numbers could be from different splits of configurations, and are not guaranteed to agree with each other.

crisp/__main__.py
thedataking: yes, added a `w.test(n_postprocess_code, n_c_code)` gate and updated comment.

crisp/__main__.py
```diff
@@ -253,6 +254,25 @@ def do_main(args, cfg):
         return
     w.accept(n_code, ('main', 'transpile'))
 
+    # Postprocess can fail by the tool itself failing, the output being not
+    # buildable, or the output failing the tests.  In all cases, failure is not
+    # fatal.
+    try:
+        n_postprocess_code = w.postprocess(n_code)
+    except (CrispError, KeyError, FileNotFoundError, toml.TomlDecodeError) as e:
```
spernsteiner: Which of these errors can actually occur?

The idea with CrispError is that it represents expected errors, where the correct behavior is to retry, and if any other error makes it to the top level, that indicates a bug in CRISP.  So my inclination is to say that this should catch only `CrispError`, and if there are specific places in `Workflow.postprocess` that are expected to produce `KeyError`, `TomlError`, etc., under certain circumstances, we should catch those and convert them to `CrispError`, similar to what we do [here](https://github.com/GaloisInc/Tractor-Crisp/blob/96900fcfb836d67b9d15d02cfa69ec0f73aea88b/crisp/llm.py#L55-L59).

spernsteiner (approved):
Looks good.  Since this falls back to doing nothing, it should be pretty safe, even if there are situations where the postprocessor fails.  If there's more work needed to make postprocess and Hayroll work together, we can do that as a separate PR.

crisp/__main__.py
thedataking: Good point, fixed!

thedataking (approved):
Approved since Hayroll integration can be done in a separate PR. Let's land this.

PR #139, opened 2026-06-02
test_eval.py: infer system_libs from cmake link flags

spernsteiner:
This was previously hardcoded to always link `libcrypto`, which is required by P01.  There's some talk of P02 possibly requiring other libraries, such as zlib, so this branch adds proper autodetection using the CMake API.

### DIFF
--- a/scripts/test_eval.py
+++ b/scripts/test_eval.py
@@ -395,7 +395,24 @@ def main():
             "this script can't handle a project with " \
             f'{num_binaries} binaries and {num_libraries} libraries'
 
-    def cmake_artifact(name, bin_main = None):
+    def cmake_artifact(target, bin_main = None):
+        name = target['name']
+
+        system_libs = set()
+        # `link` is only present for binaries and shared libraries, but not for
+        # static libraries, which have no link step.  `commandFragments` is
+        # only present "when fragments of the link command line invocation are
+        # available" (unclear when exactly that is or isn't true).  See
+        # cmake-file-api(7) for details.
+        fragments = target.get('link', {}).get('commandFragments', ())
+        for frag in fragments:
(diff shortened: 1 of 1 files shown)


ahomescu (approved):

scripts/test_eval.py
```diff
@@ -395,7 +395,19 @@ def is_library(t):
             "this script can't handle a project with " \
             f'{num_binaries} binaries and {num_libraries} libraries'
 
-    def cmake_artifact(name, bin_main = None):
+    def cmake_artifact(target, bin_main = None):
+        name = target['name']
+
+        system_libs = set()
+        for frag in target['link']['commandFragments']:
```
thedataking: According to the [Cmake docs](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#object-kind-codemodel),  `link` is optional and only present for:

 >  non-imported executables and shared library targets that link into a runtime binary

I guess that means shared libraries won't have this attribute. Is that a problem?

spernsteiner: Sounds like static libraries won't have it, but shared libraries will, which makes sense.  From what I've heard, P02 should have a binary and a shared library, but no static libraries, so we should be fine.  I'll add a comment for future reference though.

thedataking: I think you need to use `archive` for static library targets instead. It too has a `commandFragments` field.

So we need to handle shared and static libraries differently AND we need to handle cases where `commandFragments` is not present. could do something like:

```python
link_or_archive = target.get('link', target.get('archive'))
for frag in (link_or_archive or {}).get('commandFragments', []):
   # ...
```

thedataking (approved):

PR #148, opened 2026-06-20
Skip anonymous items

brk:
This would otherwise cause panics due to name collisions in code which (say, due to `bytemuck` derives) ends up with multiple anonymous const items.

Note that `_` is a valid identifier in C but not a valid function name in Rust.

### DIFF
--- a/tools/find_unsafe2/src/lib.rs
+++ b/tools/find_unsafe2/src/lib.rs
@@ -442,10 +442,19 @@ pub fn process(tcx: TyCtxt) -> Outputs {
 
     for item in items {
         if let Some(body) = item.body() {
+            let key: String = item.name();
+
+            // Skip anonymous items (`const _: () = { ... }`) and anything nested inside them (e.g.
+            // closures).  These can be emitted by derive macros (like bytemuck's `Pod`/`Zeroable`
+            // compile-time assertions) and all share the path segment `_`, so they can't be
+            // uniquely keyed by name.
+            if key == "_" || key.ends_with("::_") || key.contains("::_::") {
+                continue;
+            }
+
             let mut v = FunctionVisitor::new(&body);
             v.visit_body(&body);
 
-            let key: String = item.name();
             let mut value = FunctionOutputs {
                 total_unsafe: 0,    // Calculated later
                 is_unsafe_fn: is_unsafe_fn(item),


thedataking (approved):
LGTM. 

The failing `docker-build` workflow seems unrelated to this change. Looks like  an infra error.

brk:
We see spurious failure rates of 10-15% on our GitHub Actions runs, usually due to either networking issues or under-provisioned disks.

tools/find_unsafe2/src/lib.rs
```diff
@@ -442,10 +442,19 @@ pub fn process(tcx: TyCtxt) -> Outputs {
 
     for item in items {
         if let Some(body) = item.body() {
+            let key: String = item.name();
+
+            // Skip anonymous items (`const _: () = { ... }`) and anything nested inside them (e.g.
```
spernsteiner, line 447: My worry with skipping code like this is it gives the agent a place to hide unsafe operations.  For example ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2be85f6fab7363a5aca97456c970e01a)):

```Rust
struct Ptr(*mut i32);

const _: () = {
    impl Ptr {
        pub fn get(&self) -> &i32 {
            unsafe { &*self.0 }
        }
    }
    ()
};
```

I think it would be better to process the nested items like normal, but adjust the `key` to disambiguate.  Appending the `DefId` would probably work for this, even though it's not stable across code changes, since the items that need disambiguation also happen to be fully safe (if the `check-unsafe2` fails to find an entry in the JSON because the `DefId` changed, it will assume the function is new and require it to be fully safe, which is fine in this case).

Note there's a similar ambiguity issue with nested functions like this:
```Rust
fn f() {
    { fn g() {} }
    { fn g() {} }
}
```
But the `const _` pattern is much more common, since it's often used when implementing derive macros.

spernsteiner:
CI is failing because this is a PR from an external user, so the CI can't access the Github Actions secrets to get the access token for our private fork of Test-Corpus.  I'm not sure if there's a way to bless the PR to allow the CI to use secrets, but either way, based on what code this touches, I think it's fine to merge even without running the `docker-build` CI job (which is the only one that needs Test-Corpus).

tools/find_unsafe2/src/lib.rs
brk: Oh, yikes! This approach works for Tenjin's needs, since we can implicitly treat any anonymous const items as trusted. But it does seem much more questionable in the context of CRISP.

PR #150, opened 2026-07-02
WIP: Support resuming Codex agent session between turns

randomPoison:
> NOTE: This PR is pure Codex output, I have written none of this by hand. I am still testing this out, and will not cleanup the code until after doing some test runs to figure out if this change is even useful.
>
> ~Also, this branch currently includes an unrelated change to fix the check-unsafe2 cargo command. Once I have tested that in a long-running CRISP session I'll split that off into its own PR.~

Add a couple of options for resuming the Codex session between turns of the CRISP loop:

- `--resume-codex-session={always,rejected,never}` - Specify when to resume the session:
  - `always` - Always resume, i.e. use one session for the entire translation.
  - `rejected` - Resume only on rejected changes to give the agent a chance to fix errors, but use a fresh session after a change is accepted.
  - `never` - Never resume, so use a fresh session for every turn of the loop. This is the current behavior on master.
- `--resume-prompt={short,full}` to either use a shortened prompt when resuming or just repeat the full prompt every time. I want to A/B test these two to see if we can observe any change in performance after context compaction is happening. My assumption is that if we don't repeat our instructions in the prompt, then the full context will be lost after compaction and we'll see the agent going off the rails more.

### DIFF
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -77,6 +77,10 @@ def parse_args():
     main.add_argument('--persist-codex-session', action='store_true',
         help='resume the previous Codex session on each agent turn; '
             'requires --llm-mode=agent or similar')
+    main.add_argument('--resume-prompt',
+        choices=('short', 'full'),
+        default='short',
+        help='prompt to use when resuming a persisted Codex session')
 
     safety_loop = sub.add_parser('safety-loop')
     safety_loop.add_argument('--c-code', default='c_code')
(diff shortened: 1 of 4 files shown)


spernsteiner:
I skimmed this briefly and one thing that stood out is the new arguments in `Workflow.do_safety_step_agent`.  The original purpose of the `Workflow` type was to reduce the number of arguments required (compared to directly calling `llm`/`agent`/`analysis` functions) so that it's easier to experiment with different combinations and orders of steps.  It seems like we now have a lot of things that have to be threaded through several steps manually, which is exactly the sort of thing that `Workflow` was meant to prevent.  I wonder if we ought to move some of the state into `Workflow` (or maybe a new wrapper around `Workflow`), or bundle up state into a dataclass so `__main__` only has one thing to thread around, or something like that.  Another idea would be to move "static" stuff that almost never changes into a field or into a wrapper, so we could do something like `w.set_c_code(n_c_code)` or `w = w.with_c_code(n_c_code)` and have `c_code` filled in automatically for any step that needs it.

randomPoison:
After running this against cJSON it looks like resuming the Codex session every turn massively increases token usage without meaningfully improving performance. I'm going to say let's abandon this for now. It may at some point make sense to resume previous Codex sessions, maybe for the planning agent or to have the agent fix a broken change that it submitted, but for now I think this is a dead end.

PR #151, opened 2026-07-06
Use fresh target dir when running `cargo check-unsafe2`

randomPoison:
> NOTE: I am still testing this, but I'm putting the code up early for visiblity/discussion.

The check-unsafe2 Cargo command works by running `cargo build` with `RUSTC_WRAPPER` set to the `check_unsafe2` binary. This means that if the project was already built before running `check-unsafe2`, `cargo build` will be a no-op and don't perform the unsafe count. This then results in failed edits in the CRISP loop when check-unsafe2 fails to report increased unsafe counts to the agent.

~This PR attempts to fix this by automatically running `cargo clean` before `cargo build`, ensuring that Cargo will actually build the project. There's a problem with this solution, though: If the agent runs `cargo build` and `cargo check-unsafe2` at the same time, then we can potentially rebuild the project between the `cargo clean` and `cargo build` steps done by check-unsafe2, causing the same bug to still happen. To work around that, I've added a note to the agent prompt that tells it to not run any other Cargo commands at the same time as `cargo check-unsafe2`. That doesn't feel like a good solution to me, but I don't know if there's a good way for us to get Cargo to always re-run even when the workspace is clean.~

This PR fixes this by using a fresh `target` dir for the `cargo build` command run by `cargo check-unsafe2`. This ensures a full build every time and prevents any concurrent `cargo build` operations in the workspace from interfering with the unsafe checker.

### DIFF
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -154,8 +154,6 @@ Your changes must not introduce new unsafe code within implementation functions.
 cargo check-unsafe2 --manifest-path {cargo_dir_path}/Cargo.toml
 ```
 This will report an error for any unsafe code that was improperly added during your edits. It also reports errors on any newly added "unsafe-adjacent" code, including int-to-pointer casts and arguments or fields of raw pointer type.
-
-NOTE: Do not run `cargo build` or any other `cargo` command at the same time as `check-unsafe2`, as this can cause `check-unsafe2` to fail to report increased unsafe counts.
 '''
 
 AGENT_AFTER_REFACTORING_RUN_TESTS = '''
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -154,6 +154,8 @@ Your changes must not introduce new unsafe code within implementation functions.
 cargo check-unsafe2 --manifest-path {cargo_dir_path}/Cargo.toml
 ```
 This will report an error for any unsafe code that was improperly added during your edits. It also reports errors on any newly added "unsafe-adjacent" code, including int-to-pointer casts and arguments or fields of raw pointer type.
+
+NOTE: Do not run `cargo build` or any other `cargo` command at the same time as `check-unsafe2`, as this can cause `check-unsafe2` to fail to report increased unsafe counts.
 '''
 
 AGENT_AFTER_REFACTORING_RUN_TESTS = '''
--- a/tools/find_unsafe2/cargo_subcommands/src/lib.rs
+++ b/tools/find_unsafe2/cargo_subcommands/src/lib.rs
@@ -16,6 +16,27 @@ fn rustc_print_sysroot(opt_toolchain: Option<&str>) -> String {
     String::from_utf8(output.stdout).unwrap()
 }
 
+fn cargo_cmd(opt_toolchain: Option<&str>) -> Command {
+    let mut cmd = Command::new("cargo");
+    if let Some(toolchain) = opt_toolchain {
+        cmd.arg(format!("+{toolchain}"));
+    }
+    cmd
+}
+
+fn manifest_path_arg() -> Option<String> {
+    let mut args = env::args().skip(2);
+    while let Some(arg) = args.next() {
+        if arg == "--manifest-path" {
+            return args.next();
+        }
+        if let Some(value) = arg.strip_prefix("--manifest-path=") {
+            return Some(value.to_owned());
+        }
+    }
+    None
+}
+
 pub fn cargo_subcommand_main(wrapper_exe: &Path) -> ! {
     let opt_toolchain = option_env!("RUSTUP_TOOLCHAIN");
 
@@ -50,10 +71,22 @@ pub fn cargo_subcommand_main(wrapper_exe: &Path) -> ! {
     // different toolchain from the one `find_unsafe2` was built with.  If the parent toolchain is
     // too far apart in version, its `cargo` might be incompatible with `find_unsafe2`'s wrapped
     // `rustc`.
-    let mut cmd = Command::new("cargo");
-    if let Some(toolchain) = opt_toolchain {
-        cmd.arg(format!("+{toolchain}"));
+    let mut clean_cmd = cargo_cmd(opt_toolchain);
+    clean_cmd.arg("clean");
+    if let Some(manifest_path) = manifest_path_arg() {
+        clean_cmd.arg("--manifest-path").arg(manifest_path);
     }
+    clean_cmd.arg("--workspace");
+    eprintln!("exec: {clean_cmd:?}");
+    let status = clean_cmd.status().unwrap();
+    assert!(
+        status.success(),
+        "{:?} exited with code {:?}",
+        clean_cmd,
+        status
+    );
+
+    let mut cmd = cargo_cmd(opt_toolchain);
     cmd.arg("build")
         .args(env::args().skip(2))
         .env(LIB_PATH_VAR, new_lib_path)
--- a/tools/find_unsafe2/cargo_subcommands/src/lib.rs
+++ b/tools/find_unsafe2/cargo_subcommands/src/lib.rs
@@ -1,7 +1,6 @@
 use std::env;
 use std::path::{self, Path};
 use std::process::{self, Command};
-use std::os::unix::process::CommandExt;
 
 fn rustc_print_sysroot(opt_toolchain: Option<&str>) -> String {
     let mut cmd = Command::new("rustc");
@@ -16,7 +15,7 @@ fn rustc_print_sysroot(opt_toolchain: Option<&str>) -> String {
     String::from_utf8(output.stdout).unwrap()
 }
 
-pub fn cargo_subcommand_main(wrapper_exe: &Path) -> ! {
+pub fn cargo_subcommand_main(wrapper_exe: &Path) {
     let opt_toolchain = option_env!("RUSTUP_TOOLCHAIN");
 
     // LD_LIBRARY_PATH handling
@@ -78,6 +77,10 @@ pub fn cargo_subcommand_main(wrapper_exe: &Path) -> ! {
         .env(SRC_DIR_VAR, src_dir_abs)
         .env(JSON_DIR_VAR, json_dir_abs);
     eprintln!("exec: {cmd:?}");
-    let err = cmd.exec();
-    panic!("exec failed: {:?}", err);
+    let status = cmd
+        .status()
+        .expect("Failed to run `cargo build` with RUSTC_WRAPPER");
+    if !status.success() {
+        panic!("Failed to run {}: {status}", wrapper_exe.display())
+    }
 }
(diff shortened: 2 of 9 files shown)


tools/find_unsafe2/cargo_subcommands/src/lib.rs
```diff
@@ -50,10 +71,22 @@ pub fn cargo_subcommand_main(wrapper_exe: &Path) -> ! {
...
+    let status = clean_cmd.status().unwrap();
+    assert!(
+        status.success(),
+        "{:?} exited with code {:?}",
+        clean_cmd,
+        status
+    );
+
+    let mut cmd = cargo_cmd(opt_toolchain);
     cmd.arg("build")
```
ahomescu, line 71: Alternatively, why don't we just pass in a new random `--target-dir /tmp/tmpXYZ` here so it always rebuilds from scratch?

randomPoison: Ah yeah, that could work! Now that I think about it, that was Codex's initial attempt at a fix. I rejected it because it seemed weird to create a fresh target dir, but now that I understand the issue better, that seems like it would work. I'll give that a try.

spernsteiner: The downside of that is it would rebuild the dependencies every time, which could be nontrivial (if it has bitfields, that includes building `syn`)

crisp/workflow.py
```diff
@@ -154,6 +154,8 @@
 cargo check-unsafe2 --manifest-path {cargo_dir_path}/Cargo.toml
 ```
 This will report an error for any unsafe code that was improperly added during your edits. It also reports errors on any newly added "unsafe-adjacent" code, including int-to-pointer casts and arguments or fields of raw pointer type.
+
+NOTE: Do not run `cargo build` or any other `cargo` command at the same time as `check-unsafe2`, as this can cause `check-unsafe2` to fail to report increased unsafe counts.
```
spernsteiner: Could we have `cargo check-unsafe2` use a separate target dir so these won't conflict?  (But the same target dir each time, so it doesn't need to rebuild deps)

tools/find_unsafe2/cargo_subcommands/tests/example.rs
```diff
@@ -0,0 +1,65 @@
...
+        .current_dir(&fixture_dir)
+        .arg("check-unsafe2")
+        .args(["--manifest-path", "example-new/Cargo.toml"])
+        .env("FIND_UNSAFE2_JSON_DIR", &json_dir)
+        .stderr(Stdio::inherit())
+        .output()
+        .unwrap();
+    assert!(
+        !output.status.success(),
+        "cargo-check-unsafe2 succeeded unexpectedly"
```
spernsteiner, line 67: Can you add some comments in here saying what we expect each command to do?  I had to read this test pretty carefully to understand what it's actually trying to do

ahomescu:
```
[19:55:39    76] codex
[19:55:39    76] The unsafe JSON was stale despite `cargo find-unsafe2` saying it finished instantly; it still showed the old Adler pointer loop. I’m following the scratchpad warning here: clean the Rust target and rerun the unsafe finder so the report reflects the edited source.
[19:55:40    77] exec
[19:55:40    77] /bin/bash -lc 'cargo clean --manifest-path translated_rust/Cargo.toml && cargo find-unsafe2 --manifest-path translated_rust/Cargo.toml' in /root/work
```

Funny enough, the agent sometimes figures it out.

tools/find_unsafe2/cargo_subcommands/src/lib.rs
randomPoison: Yeah, that was a potential downside of this. That's why I have it do `cargo clean --workspace`, so it doesn't fully clean all dependencies. But also, we replace rustc with our own binary, so are we actually doing any kind of codegen when we run check-unsafe? We have Cargo act as if its doing a full build, but if we aren't actually running rustc then it probably doesn't matter.

spernsteiner: > are we actually doing any kind of codegen when we run check-unsafe?

Yes, it runs all the normal codegen in addition to the check-unsafe logic.  Otherwise it wouldn't work after `cargo clean` because the later steps wouldn't be able to find `.rlib`s/`.rmeta`s for their dependencies.

tools/find_unsafe2/cargo_subcommands/tests/example.rs
randomPoison: Sure, I added some comments. Let me know if anything is still unclear.

randomPoison:
I've updated this PR to use a fresh `target` dir every time, which fixes the problem better than doing `cargo clean` does. This should _mostly_ fix the issues with check-unsafe2 failing to report increased unsafe counts, which greatly reduces the number of rejected edits. There are still some secondary issues where the agent might accidentally overwrite the unsafe JSON, which can then lead to rejected edits, but I'll put up a different PR with prompt changes that give the agent more context/advice on how to handle that stuff.

tools/find_unsafe2/cargo_subcommands/src/lib.rs
```diff
@@ -46,6 +48,26 @@ pub fn cargo_subcommand_main(wrapper_exe: &Path) -> ! {
...
+    // dir, forcing a full build every time.
+    let timestamp_nanos = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_nanos();
+    let target_dir = env::temp_dir().join(format!(
+        "find_unsafe2-target-{}-{timestamp_nanos}",
+        process::id()
+    ));
+    fs::create_dir_all(&target_dir).unwrap();
```
ahomescu: Does this directory get deleted when it's not needed?

tools/find_unsafe2/cargo_subcommands/src/lib.rs
```diff
@@ -46,6 +48,26 @@ pub fn cargo_subcommand_main(wrapper_exe: &Path) -> ! {
...
+    // rustc wrapper.
+    //
+    // If we try to run `cargo build` within the normal workspace and the workspace
+    // is clean (i.e. no changes have been made since the last `cargo build`), then
+    // `cargo build` is a no-op and never actually invokes our wrapper. This means
+    // that if the agent does `cargo build` right before (or at the same time as)
+    // `cargo check-unsafe2` then we won't report any increased unsafe counts. To
+    // ensure that Cargo always builds the project, we use a temp dir as the target
+    // dir, forcing a full build every time.
+    let timestamp_nanos = SystemTime::now()
```
ahomescu: It might be more robust to use an existing crate like https://docs.rs/tempdir which handles both the unique naming and automatic deletion. I thought the standard library would have this functionality (like in Python), but I guess not.

tools/find_unsafe2/cargo_subcommands/tests/example.rs
```diff
@@ -0,0 +1,81 @@
+use std::process::{Command, Stdio};
+
+// HACK: Clear RUSTFLAGS when running the build/tools on the unsafe code. When
```
ahomescu: Are there no flags we _do_ want to pass in though? It might be better to just scrub out `-D warnings`.

tools/find_unsafe2/cargo_subcommands/src/lib.rs
randomPoison: No it does not, but I figured that's not a problem because each turn of the loop we're starting a fresh docker instance, right?

I looked at using [tempfile](https://crates.io/crates/tempfile) to automatically cleanup the temp dir, but we can't do that with how we're currently exectuing the command with the `exec` function. [`exec`](https://doc.rust-lang.org/stable/std/os/unix/process/trait.CommandExt.html#tymethod.exec) basically aborts the process on success, meaning that we can't do any cleanup after the `cargo build` completes, including using tempfile's destructor-based cleanup. @spernsteiner do you know why we're doing `exec` here instead of something more normal like `status`? Do we need the script to abort on success for some reason?

tools/find_unsafe2/cargo_subcommands/src/lib.rs
randomPoison: tempdir is unmaintained, but looks like https://crates.io/crates/tempfile exists and does what we need.

tools/find_unsafe2/cargo_subcommands/tests/example.rs
randomPoison: I'm not sure why we'd want to pass RUSTFLAGS through to the test compilation, mostly it seems to just interfere with running the test.

ahomescu (approved):
LGTM, handing it over to @spernsteiner.

tools/find_unsafe2/cargo_subcommands/src/lib.rs
spernsteiner: I just default to `exec` when "tail-calling" another binary because it's slightly lower overhead.  If we now have cleanup to do after the subprocess ends, it should be fine to switch it to run normally instead.

tools/find_unsafe2/cargo_subcommands/src/lib.rs
```diff
@@ -57,6 +72,7 @@ pub fn cargo_subcommand_main(wrapper_exe: &Path) -> ! {
     cmd.arg("build")
         .args(env::args().skip(2))
         .env(LIB_PATH_VAR, new_lib_path)
+        .env("CARGO_TARGET_DIR", target_dir.path())
```
spernsteiner: Maybe it would be better to pass `--target-dir foo` right after `build`, before `env::args()`?  It looks like `--target-dir` takes precedence over `$CARGO_TARGET_DIR`, so if the user for some reason sets their own `--target-dir`, our custom target dir would be silently ignored if we use an env var, but would result in an explicit error if we pass our own `--target-dir` option.

tools/find_unsafe2/cargo_subcommands/tests/example.rs
```diff
@@ -0,0 +1,81 @@
...
+// running in CI we use `-D warnings`, which then leaks into the build of the
+// unsafe test code causing build errors. This helper clears that flag so that
+// we can allow warnings in the test code.
```
spernsteiner: Should we just fix the warnings in the example crate instead?  I don't think it would cause any problems for the tests.  I just didn't bother initially since the code didn't need to build in CI or actually work at all.

spernsteiner (approved):
Just a few minor comments

tools/find_unsafe2/cargo_subcommands/src/lib.rs
randomPoison: Okay, I switched it to use `status` instead of `exec`, and I have it checking the exit status and panicking if the `cargo build` returned an error code. This should allow it to clean up the temp dir before exiting.

randomPoison:
@spernsteiner I've addressed your feedback and re-tested with the updated functionality and seems like everything is working. This should be ready to merge if it looks good to you.

spernsteiner (approved):

PR #152, opened 2026-07-06
Split agent planning into a separate step

ahomescu:
Split the planning into a separate part outside of the main safety loop, instead of the old approach of prompting to create a `SAFETY_PLAN.md` file if it doesn't already exist.

### DIFF
--- a/crisp/agent.py
+++ b/crisp/agent.py
@@ -24,7 +24,8 @@ def _snapshot_to_family_alias(model: str) -> str:
     m = _SNAPSHOT_SUFFIX.match(model)
     return m.group("alias") if m else model
 
-def _codex_command(cfg: Config, subcmd: str, args: list[str], codex_login: bool = False) -> list[str]:
+def _codex_command(cfg: Config, subcmd: str, args: list[str],
+                   model: str, codex_login: bool = False) -> list[str]:
     cmd = ['codex', subcmd]
 
     if codex_login:
@@ -32,7 +33,7 @@ def _codex_command(cfg: Config, subcmd: str, args: list[str], codex_login: bool 
         # override the model; everything else uses codex's defaults.
         # The --model flag does not support snapshot-style model identifiers so
         # we attempt to convert snapshots to model family aliases.
-        model = _snapshot_to_family_alias(llm.API_MODEL or cfg.models.agent)
+        model = _snapshot_to_family_alias(llm.API_MODEL or model)
         cmd += ['--model', model]
     else:
         config_settings = {
@@ -41,7 +42,7 @@ def _codex_command(cfg: Config, subcmd: str, args: list[str], codex_login: bool 
             #'model_providers.crisp.api_key': llm.API_KEY or 'sk-no-api-key',
             'model_providers.crisp.env_key': 'CRISP_API_KEY',
             'model_provider': 'crisp',
-            'model': llm.API_MODEL or cfg.models.agent,
+            'model': llm.API_MODEL or model,
             # TODO: OpenAI pricing is based on input and output tokens, with 
             # long context tokens costing twice as much as short context ones.
             # We might want to set limits to avoid the long context pricing.
@@ -98,6 +99,7 @@ def run_rewrite(
     cfg: Config,
     mvir: MVIR,
     prompt: str,
+    model: str,
     input_code: TreeNode,
     extra_code: TreeNode | list[TreeNode] = [],
     planning_files: TreeNode | None = None,
@@ -168,7 +170,7 @@ def run_rewrite(
                 '--dangerously-bypass-approvals-and-sandbox',
                 '--skip-git-repo-check',
                 prompt,
-            ], codex_login=codex_login)
+            ], codex_login=codex_login, model=model)
             print(codex_cmd)
             all_cmds = [mkdir_codex, codex_cmd] + clean_cmds
 
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -1310,6 +1310,13 @@ class Workflow:
             self.find_unsafe2_json(n_code),
         ]
         if n_test_code is not None:
+            # In planning mode, we do provide the entire C code to the planner
+            # since that might increase the quality of the plan. However, if
+            # we later run the safety loop in `agent_sim_no_tests` mode the
+            # safety agent will not have access to the C code, and the plan
+            # might contain references to it. This might confuse the agent and
+            # break the `agent_sim_no_tests` mode.
+            # TODO: un-break that mode somehow (without hiding the C code).
             extra_code.append(n_test_code)
 
         prompt = AGENT_PLAN_PROMPT.format(
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -1197,7 +1197,7 @@ class Workflow:
         )
         if prompt_suffix is not None:
             prompt = f'{prompt}\n\n{prompt_suffix}'
-        return agent.run_rewrite(cfg, mvir, prompt, n_code,
+        return agent.run_rewrite(cfg, mvir, prompt, self.cfg.models.agent_loop, n_code,
             extra_code = extra_code,
             planning_files = n_plans,
             codex_login=self.codex_login,
@@ -1316,7 +1316,7 @@ class Workflow:
             cargo_dir_path = cargo_dir,
             after_refactoring_instruction = after_refactoring_instruction,
         )
-        return agent.run_rewrite(cfg, mvir, prompt, n_code,
+        return agent.run_rewrite(cfg, mvir, prompt, self.cfg.models.agent_plan, n_code,
             extra_code = extra_code,
             planning_files = None,
             codex_login=self.codex_login,
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -118,21 +118,9 @@ New signatures:
 {input_files}
 '''
 
-AGENT_SAFETY_PROMPT = '''
-Please refactor the Rust code in `{cargo_dir_path}` to be safe, without changing its behavior.  The goal is to make the code fully safe, including migrating to safe memory management (`Box`/`Vec`/`Rc`) and  safe pointer types (`&T`/`&[T]`) throughout.
-
-{target_goal}
-
-You are executing one iteration of a loop that will be re-invoked on the same codebase until the unsafe count reaches zero or progress stalls. To carry state between iterations, maintain a planning file at `SAFETY_PLAN.md`:
-
-- **First, check whether `SAFETY_PLAN.md` already exists.** If it does, read it in full. It is your own notes from prior iterations.
-- If the existing `SAFETY_PLAN.md` applies to your current target, Use it to pick up where the previous run left off rather than starting over.
-- If `SAFETY_PLAN.md` does not exist, or if it exists but applies to an unrelated target, make a new plan for addressing the current target. Examine the target along with any callers/callees, use sites, and other related logic to understand the functionality it provides. Think about how to implement the same functionality using only safe Rust constructs. Then, plan out step by step how to migrate from the unsafe version to the safe equivalent. This plan may be as small as a single step if the target is simple, or it may be many steps if the target is complex or has many interactions. Once you have a plan, write it to `SAFETY_PLAN.md`.
-- **Before you finish, update `SAFETY_PLAN.md`** to reflect what you actually did this iteration, what is now complete, what remains, and any pitfalls or dead ends future iterations should avoid. Keep it concise — it is a working scratchpad, not a report.
-- When rewriting the plan due to switching targets, pay close attention to any relevant pitfalls or dead ends recorded in the previous plan such that future iterations do not repeat the same mistakes. This is also a good time to trim and clean up the plan so it only contains the next steps and relevant notes.
-- If a planned unit turns out to be too complex or blocked by prerequisite work, update `SAFETY_PLAN.md` to record the blocker, split the work into smaller steps, and switch to the prerequisite or smaller step. Prefer steps that directly reduce the unsafe count, but preliminary safe refactors are acceptable when they are necessary to remove unsafe code in a later iteration.
-
-Then carry out the next step of the plan.
+AGENT_PLAN_PROMPT = '''
+Make a plan to refactor the Rust code in `{cargo_dir_path}` to be safe, without changing its behavior.  The goal is to make the code fully safe, including migrating to safe memory management (`Box`/`Vec`/`Rc`) and  safe pointer types (`&T`/`&[T]`) throughout.
+Write the plan to a file named `SAFETY_PLAN.md`.
 
 The code contains two kinds of functions: implementation functions and FFI entry points. A function is an FFI entry point if it has the `#[no_mangle]` or `#[export_name]` attribute ("export attributes"). Note that just having `unsafe extern "C"` qualifiers without export attributes does NOT make a function an FFI entry point. All functions without export attributes are implementation functions.
 
@@ -142,10 +130,25 @@ Implementation code may be freely refactored to improve safety. Here are some ex
 - Eliminate calls to FFI imports and other unsafe functions. Try to replace these with suitable safe Rust operations, e.g. `printf` -> `println!`.
 - Remove the `unsafe` qualifier from functions that no longer need it, so that their callers can be made safe.
 
-For FFI entry points, the following rules apply:
+For FFI entry points, the following rules apply (these rules must be copied verbatim into the plan and never changed or broken):
 - You must not change the signature. FFI entry point signatures must remain exactly as-is to ensure ABI compatibility with the current version of the code. Don't remove `unsafe` or `extern "C"` qualifiers from FFI entry points.
 - For struct types that appear only behind a pointer, you may assume the struct is opaque to the user of the library (unless otherwise indicated within the code itself), as is considered best practice in C. This means the struct layout is not part of the ABI, so you may freely change the field types to improve safety.
 - Each FFI entry point should convert the inputs from unsafe types to safe ones (e.g. `*const T` -> `&T`) if needed, dispatch to an implementation function, and convert the results back to unsafe types if needed. Do not add extraneous unsafe code to FFI entry points.
+
+{after_refactoring_instruction}
+
+Your changes must not introduce new unsafe code within implementation functions. You can check your work using this command:
+```sh
+cargo check-unsafe2 --manifest-path {cargo_dir_path}/Cargo.toml
+```
+This will report an error for any unsafe code that was improperly added during your edits. It also reports errors on any newly added "unsafe-adjacent" code, including int-to-pointer casts and arguments or fields of raw pointer type.
+'''
+
+AGENT_SAFETY_PROMPT = '''
+Continue the plan from `SAFETY_PLAN.md`.
+**Before you finish, update `SAFETY_PLAN.md`** to reflect what you actually did this iteration, what is now complete, what remains, and any pitfalls or dead ends future iterations should avoid. Keep it concise — it is a working scratchpad, not a report.
+
+{target_goal}
 
 {after_refactoring_instruction}
 
@@ -1288,6 +1291,41 @@ class Workflow:
             n_new_code = self.llm_safety(n_code)
 
         return self.do_validate_and_repair(n_code, n_new_code, n_test_code)
+
+    @step
+    def do_safety_plan_agent(
+        self,
+        n_code: TreeNode,
+        n_test_code: TreeNode,
+    ) -> tuple[TreeNode | None, TreeNode | None]:
+        cfg, mvir = self.cfg, self.mvir
+        cargo_dir = cfg.relative_path(cfg.transpile.output_dir)
+
+        if cfg.test_command is not None:
+            after_refactoring_instruction = AGENT_AFTER_REFACTORING_RUN_TESTS \
+                    .format(test_cmd = cfg.test_command)
+        else:
+            after_refactoring_instruction = AGENT_AFTER_REFACTORING_BUILD
+
+        extra_code = [
+            self.find_unsafe2_json(n_code),
+        ]
+        if n_test_code is not None:
+            extra_code.append(n_test_code)
+
+        prompt = AGENT_PLAN_PROMPT.format(
+            cargo_dir_path = cargo_dir,
+            after_refactoring_instruction = after_refactoring_instruction,
+        )
+        return agent.run_rewrite(cfg, mvir, prompt, n_code,
+            extra_code = extra_code,
+            planning_files = None,
+            codex_login=self.codex_login,
+            clean_cmds = [
+                ['cargo', 'clean', '--manifest-path', os.path.join(cargo_dir, 'Cargo.toml')],
+            ],
+            find_unsafe2_json_dir = analysis.UNSAFE_JSON_DIR,
+        )
 
     @step
     def do_safety_step_agent(
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -178,10 +178,6 @@ AGENT_TARGET_GOAL_FUNCTION = '''
 Your current target is the `{func_name}` function, along with any similar or closely related functions.  Your goal is to change the function in question to use only safe types (such as `Box`, `Vec`, or `Rc`) internally and in its signature, and to replace any unsafe FFI calls (e.g. `libc::printf`) with safe equivalents.  You should aim to make callees of the target function safe if it's feasible to do so, that way the entire call tree is safe.
 '''.strip()
 
-AGENT_TARGET_GOAL_OTHER = '''
-You currently have no specific target.  Scan the codebase for any remaining unsafe types, operations, or definitions (excluding FFI entry points) and identify a single reasonably-scoped unit of code (such as a file/module, a data structure and its related functions, or even a set of related struct fields) that uses `unsafe` for you to work on.
-'''.strip()
-
 class AgentTarget:
     def prompt(self):
         return self.PROMPT_FMT.format(**dataclasses.asdict(self))
@@ -208,7 +204,10 @@ class AgentTargetOther(AgentTarget):
     """
     Remove any leftover unsafe from the codebase.
     """
-    PROMPT_FMT = AGENT_TARGET_GOAL_OTHER
+    # The prompt asks the agent to continue the plan, so we
+    # expect it to make a reasonable choice for the next target
+    # based on that document.
+    PROMPT_FMT = ""
 
 
 _CRISP_DIR = os.path.dirname(os.path.dirname(__file__))
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ cd /root/project
 export CRISP_API_BASE=https://api.openai.com/v1
 export CRISP_API_KEY=sk-your-api-key-here
 
-# Optional: override the `models.agent`, `models.rewriter`, etc. selections
+# Optional: override the `models.agent_plan`, `models.agent_loop`, `models.rewriter`, etc. selections
 # in crisp.toml for this run.
 #export CRISP_API_MODEL=gpt-5.6-sol
 
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -371,7 +371,12 @@ def safety_loop_common(args, cfg, mvir, w, n_code, n_c_code):
 
     best_unsafe_count = None
     consecutive_failures = 0
-    n_plans = prior_agent_plans(mvir, n_code) or TreeNode.new(mvir, files={})
+    n_plans = prior_agent_plans(mvir, n_code)
+    if not n_plans:
+        if 'agent' in args.llm_mode:
+            n_plans = w.do_safety_plan_agent(n_code, n_c_code)[1]
+        else:
+            n_plans = TreeNode.new(mvir, files={})
 
     target_goal = None
     target_goal_tries = 0
--- a/crisp/config.py
+++ b/crisp/config.py
@@ -40,7 +40,8 @@ class ConfigBase:
 
 @dataclass(frozen = True)
 class ModelsConfig(ConfigBase):
-    agent: str  = "gpt-5.6-terra"
+    agent_plan: str = "gpt-5.6-sol"
+    agent_loop: str = "gpt-5.6-terra"
     postprocess: str = "gpt-5.6-luna"
     # `rewriter = None` means call `/v1/models` and pick the first from the list.
     rewriter: str | None = None
--- a/crisp/git.py
+++ b/crisp/git.py
@@ -37,6 +37,7 @@ OP_NODE_KINDS = {
 }
 
 HISTORY_INDEX_KEYS = set((kind, new) for (kind, (old, new)) in OP_NODE_KINDS.items())
+HISTORY_INDEX_OLD_KEYS = set((kind, old) for (kind, (old, new)) in OP_NODE_KINDS.items())
 
 def render(mvir: MVIR, target: TreeNode) -> pygit2.Oid:
     """
@@ -49,7 +50,19 @@ def render(mvir: MVIR, target: TreeNode) -> pygit2.Oid:
         # Try to find an op that produced the last tree.
         tree = trees[-1]
         op_node_id = None
-        for ie in mvir.index(tree.node_id()):
+        index = mvir.index(tree.node_id())
+
+        # Save a set of all nodes with `old_code`
+        old_node_ids = set()
+        for ie in index:
+            if (ie.kind, ie.key) in HISTORY_INDEX_OLD_KEYS:
+                old_node_ids.add(ie.node_id)
+
+        for ie in index:
+            if ie.node_id in old_node_ids:
+                # Ignore nodes where `old_code==new_code`,
+                # such as the planning node from agent mode
+                continue
             if (ie.kind, ie.key) in HISTORY_INDEX_KEYS:
                 op_node_id = ie.node_id
                 break


ahomescu:
I tested this on zlib and here is the plan it produced:
---
# Safety Refactor Plan for `translated_rust`

## Goal

Refactor `translated_rust` so that Rust implementation code is fully safe while preserving the current zlib ABI and behavior.

The exported ABI surface must remain compatible with the current crate. Functions with `#[export_name]` or `#[no_mangle]` are FFI entry points and must keep their signatures exactly as-is. In the current tree, the exported functions are all `#[export_name]` wrappers; the `#[no_mangle]` items found in the audit are exported data symbols, not functions.

## Current Baseline

- Crate: `translated_rust`, `cdylib` named `zlib`.
- Size: about 24.5k lines of translated Rust.
- Main unsafe-heavy modules:
  - `deflate.rs`, `inflate.rs`, `trees.rs`, `inftrees.rs`, `crc32.rs`
  - gzip file API modules: `gzlib.rs`, `gzread.rs`, `gzwrite.rs`, `gzclose.rs`
  - convenience/checksum modules: `compress.rs`, `uncompr.rs`, `adler32.rs`, `zutil.rs`
- Exported functions: 106 `#[export_name]` wrappers.
- Exported data symbols: `_dist_code`, `_length_code`, `z_errmsg`, `deflate_copyright`, `inflate_copyright`.
- Existing baseline commands currently complete:
  - `cargo build --manifest-path translated_rust/Cargo.toml`
  - `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml`

## Safety Boundary Rules

1. Keep every exported function signature exactly unchanged.
2. Keep public C ABI structs in `lib.rs` stable unless proven opaque by use:
   - Treat `z_stream_s` and `gz_header_s` as ABI-visible because C callers allocate and populate them.
   - Treat `gzFile`/`gz_state` and compression/decompression internal states as opaque implementation state.
3. Move all raw pointer validation and conversion into exported wrappers.
4. Dispatch from each wrapper into safe implementation functions that use:
   - `&T`, `&mut T`, `&[u8]`, `&mut [u8]`
   - `Box<T>` for owned single states
   - `Vec<T>` for buffers and tables
   - `Rc` only if shared ownership is actually required
5. Do not introduce new unsafe code in implementation functions. Any remaining unsafe must be limited to exported FFI entry points and legacy ABI declarations that cannot be removed until the corresponding module is migrated.

## Target Architecture

### ABI Layer

Create a small boundary layer, likely in new modules such as `ffi.rs` and `abi.rs`, for ABI constants, error mapping, safe wrapper types, and safe helper functions.

Do not put raw-pointer-to-reference conversions in ordinary non-exported helper functions, since those functions would be implementation functions under the requested rules. Instead, each exported wrapper should perform its own boundary conversion:

1. check null and length invariants;
2. do the minimal pointer-to-reference or pointer-to-slice conversion inside the exported function;
3. call only safe helper functions and safe implementation functions after conversion;
4. copy updated counters or result pointers back to the ABI structs before returning.

Shared helpers in the ABI layer may operate on already-safe references, slices, `NonNull` values, or wrapper structs, but they must not contain new unsafe code unless they are themselves exported FFI entry points.

### Safe Implementation Layer

Replace translated C-style functions with safe Rust functions:

- `adler32_z_impl(adler: uLong, buf: Option<&[u8]>) -> uLong`
- `crc32_z_impl(crc: uLong, buf: Option<&[u8]>) -> uLong`
- `deflate_impl(stream: &mut StreamIo, state: &mut DeflateState, flush: c_int) -> c_int`
- `inflate_impl(stream: &mut StreamIo, state: &mut InflateState, flush: c_int) -> c_int`
- gzip functions over `GzState`, `File`, `Vec<u8>`, and safe slices.

The existing non-exported `pub unsafe extern "C" fn name(...)` functions should be renamed or replaced by safe Rust functions. Exported wrappers keep the zlib symbol name and call the safe implementation.

## Data Model Migration

### `z_stream_s`

Preserve the C layout. Do not change field types in `z_stream_s` during the first migration because C tests pass `sizeof(z_stream)` into `deflateInit_` and `inflateInit_`.

Internally, treat `z_stream_s` as an ABI carrier only. Convert it into a safe temporary view:

- input: `&[u8]` from `next_in` and `avail_in`
- output: `&mut [u8]` from `next_out` and `avail_out`
- state: checked opaque handle to `Box<DeflateState>` or `Box<InflateState>`
- counters: update `total_in`, `total_out`, `avail_in`, `avail_out`, `next_in`, `next_out` after the safe implementation returns.

### Internal States

Replace raw fields in `deflate::internal_state`, `inflate::inflate_state`, and `gzguts_h::gz_state` with safe fields where the struct is opaque:

- `pending_buf`, `window`, `prev`, `head`, `sym_buf` -> `Vec<u8>` / `Vec<Pos>`
- `pending_out` pointer -> index into `pending_buf`
- `gzhead` -> `Option<HeaderView>` or a validated boundary handle
- `lencode`, `distcode`, `next` -> table indices or borrowed slices instead of raw table pointers
- gzip `path`, `msg` -> `CString`/`String`
- gzip `in_0`, `out` -> `Vec<u8>`
- gzip file descriptor -> `std::fs::File` or `OwnedFd`

### Unions and Statics

Replace unsafe translated data forms:

- `ct_data_s` unions -> explicit safe fields or an enum only if the variant matters. Since the C union members are both `u16`, prefer a safe struct with named `u16` fields if behavior only overlays identical storage.
- `static mut` read-only tables -> `static` or `const` arrays.
- `configuration_table` -> immutable `static` with safe function items or an enum strategy dispatch.
- `inflate_mode` numeric constants -> `#[repr(u32)] enum InflateMode` internally, with conversion at the ABI boundary if needed.
- Exported data symbols must keep symbol names and layout. If they must remain mutable for ABI compatibility, isolate mutation behind wrappers and use immutable internal copies for implementation.

## Refactor Phases

### Phase 1: Freeze ABI and Add Boundary Helpers

1. Generate and save an export inventory with `nm -D translated_rust/target/debug/libzlib.so`.
2. Add tests or a checked script that verifies exported symbol names remain present.
3. Add safe ABI helper types for stream views, error mapping, and return-value handling.
4. Update one simple wrapper at a time so the exported function itself validates raw pointers, creates safe references/slices, and calls a safe implementation.

Start with functions that do not require stream state:

- `zlibVersion`, `zlibCompileFlags`, `zError`
- `adler32*`
- `crc32_combine*`
- `compressBound*`

### Phase 2: Convert Checksums and Tables

1. Rewrite `adler32.rs` to operate on `Option<&[u8]>` and iterators/indexing instead of pointer arithmetic.
2. Rewrite `crc32.rs` to use safe byte-slice iteration. Keep the same table-driven algorithm first; only optimize after tests pass.
3. Convert CRC and inflate/deflate constant tables from `static mut` to immutable `static`/`const`.
4. Keep `get_crc_table_ffi` returning the ABI pointer, but make the implementation use safe table references.

Verification after this phase:

- `cargo build --manifest-path translated_rust/Cargo.toml`
- `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml`
- full shared-library test script.

### Phase 3: Safe Huffman Tables and Tree Code

1. Convert `inftrees::inflate_table` into a safe implementation:
   - `lens: &[u16]`
   - `table: &mut [code]`
   - `bits: &mut u32`
   - `work: &mut [u16]`
   - return the number of table entries used instead of advancing a raw pointer.
2. Convert fixed tables `lenfix`, `distfix`, `_dist_code`, `_length_code`, `static_ltree`, and `static_dtree` to immutable statics.
3. Convert tree helpers in `trees.rs` to safe state references and indexed buffer access.
4. Replace `tree_desc_s` raw table pointers with safe indices or borrowed descriptors inside the implementation state.

### Phase 4: Safe Deflate State

1. Introduce `DeflateState` as the internal owned state.
2. Allocate all internal buffers with `Vec` in `deflateInit2_ffi`, then store a boxed opaque pointer in `z_stream_s.state` for ABI compatibility.
3. Convert `deflateStateCheck` into a safe state validation helper used by wrappers.
4. Migrate helpers in dependency order:
   - `slide_hash`
   - `read_buf`
   - `fill_window`
   - `lm_init`
   - `longest_match`
   - `deflate_stored`, `deflate_fast`, `deflate_slow`, `deflate_rle`, `deflate_huff`
   - `flush_pending`, `putShortMSB`
5. Replace `memcpy`/`memset` with `copy_from_slice`, `copy_within`, `fill`, and slice assignment.
6. Remove unsafe qualifiers from all non-exported deflate functions as they become safe.

Custom allocators require special handling. To avoid unsafe calls in implementation functions, exported wrappers that allocate or free state (`deflateInit*`, `deflateEnd`, `deflateCopy`) should own all interactions with C allocator callbacks. The safe implementation should receive already-owned safe state.

### Phase 5: Safe Inflate State

1. Introduce `InflateState` with owned `Vec<u8>` window and safe code-table storage.
2. Convert `inflateStateCheck` into a safe validation helper used by wrappers.
3. Convert `inflate`, `inflate_fast`, `updatewindow`, `syncsearch`, and dictionary/header helpers to operate on safe stream views and state references.
4. Replace `lencode`, `distcode`, and `next` raw pointers with table IDs, ranges, or indices into owned arrays.
5. Eagerly or wrapper-side allocate the window at the same API points that currently may allocate, so allocator callback behavior remains as close as possible to current behavior.

For `inflateBack`, callbacks are part of the FFI boundary. Refactor the algorithm into a safe stepper that consumes safe input slices and produces safe output slices. Keep raw callback invocation in the exported `inflateBack_ffi` wrapper only.

### Phase 6: Convenience Compression APIs

After deflate/inflate are safe:

1. Rewrite `compress.rs` and `uncompr.rs` as thin safe implementations over safe stream APIs.
2. FFI wrappers validate `dest`, `destLen`, `source`, and `sourceLen` pointers and convert them to slices.
3. Preserve zlib's null-pointer and zero-length behavior exactly.

### Phase 7: gzip File API

1. Replace `malloc/free` ownership in `gzlib.rs`, `gzread.rs`, `gzwrite.rs`, and `gzclose.rs` with `Box<GzState>` and `Vec<u8>`.
2. Replace libc file calls:
   - `open` -> `OpenOptions`
   - `read`/`write` -> `Read`/`Write`
   - `lseek64` -> `Seek`
   - `close` -> `Drop`
   - `errno`/`strerror` -> `std::io::Error`
3. Handle `gzdopen` in the exported wrapper by converting the raw fd into an owned Rust file handle at the boundary.
4. Convert gzip input and output buffers to slices and indices.
5. Preserve current error codes and messages returned through `gzerror`.

### Phase 8: Remove Unsafe Imports and Feature Gates

When no implementation code calls libc memory/string/file functions:

1. Delete unused imports from `stdlib`.
2. Remove `#![feature(raw_ref_op)]` and `#![feature(extern_types)]` if no longer needed.
3. Replace `ZLIB_VERSION` transmute initializers with safe byte/string constants.
4. Remove `unsafe extern "C"` from every implementation function that no longer needs it.

## Verification Plan

Run these after every phase and before merging:

```sh
cargo build --manifest-path translated_rust/Cargo.toml
cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml
```

Run the full ABI behavior test after each module-level migration:

```sh
set -e
(cd translated_rust && cargo build)
LIBZLIB_SO=$(readlink -f translated_rust/target/debug/libzlib.so)

rm -rf test_build
mkdir test_build
cd test_build
sh ../configure
make -j `nproc` shared
echo "Replacing libz.so with $LIBZLIB_SO"
cp $LIBZLIB_SO libz.so.1.3.2.1-motley
make testshared
cmp $LIBZLIB_SO libz.so.1.3.2.1-motley
```

Also run targeted symbol checks:

```sh
nm -D translated_rust/target/debug/libzlib.so | sort > /tmp/zlib-symbols.after
```

Compare this against the baseline export list. Any removed or renamed exported symbol is a blocker.

## Definition of Done

- All exported function signatures are unchanged.
- Shared-library tests pass with the Rust `libzlib.so` substituted for the C build output.
- `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml` passes.
- No implementation function contains unsafe blocks, raw pointer arguments, raw pointer fields, int-to-pointer casts, or calls to unsafe functions.
- Remaining raw pointers are isolated to ABI structs, exported wrapper signatures, and tightly audited boundary conversions.
- Internal ownership is represented with `Box`, `Vec`, safe references, and safe indices rather than `malloc`, `free`, and pointer arithmetic.

ahomescu:
Here is the plan drawn up for libexpat
---
# Safety Refactor Plan

This plan covers the Rust port in `translated_rust` only. Do not edit C sources or headers under `expat/`.

## Current State

- The crate is a C2Rust-style `cdylib` with most code in `src/xmlparse.rs`, `src/xmltok.rs`, and `src/xmlrole.rs`.
- Exported FFI entry points are exactly the functions with `#[no_mangle]` or `#[export_name]`. Baseline counts from inspection:
  - `src/xmlparse.rs`: 79 export attributes.
  - `src/xmltok.rs`: 14 export attributes.
  - `src/xmlrole.rs`: 2 export attributes.
- Many functions are `unsafe extern "C"` but are implementation functions because they have no export attribute. These may be converted to safe Rust functions once their signatures and callers are safe.
- The dominant unsafe sources are raw parser handles, pointer ranges, function pointer tables, `static mut` constants, hand-managed allocation, C library calls, and C-style intrusive linked lists.
- Baseline verification:
  - `cargo build` in `translated_rust` succeeds.
  - `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml` succeeds on the current baseline.
  - `cargo fmt --all --check` currently reports existing formatting drift in generated Rust files; any refactor commit should either include rustfmt output or clearly normalize formatting in the touched files.

## Refactor Rules

1. Preserve every FFI entry point signature exactly. Do not remove `unsafe`, `extern "C"`, exported symbol names, or argument/return types from functions with `#[no_mangle]` or `#[export_name]`.
2. Keep unsafe code at the FFI edge. Each exported function should validate nullable/raw inputs, convert to safe wrapper types, call safe implementation code, then convert results back to ABI-compatible values.
3. Do not add new unsafe code to implementation functions. When an unsafe operation is unavoidable during migration, put it in a narrow boundary helper whose purpose is to adapt FFI data to safe Rust.
4. Treat `XML_ParserStruct` and internal parser-side structs as opaque unless they are directly exposed by value through public FFI. Their fields may be changed to safe Rust ownership and reference types.
5. Preserve public C-facing `#[repr(C)]` layouts for types passed by value or embedded in public ABI structs. For internal opaque structs, prefer removing `#[repr(C)]` after all direct C layout assumptions are gone.
6. Run the provided C test harness after each behavior-changing phase, and run `cargo check-unsafe2` after each safety phase.

## Phase 1: Draw the FFI Boundary

Purpose: separate ABI shims from implementation so later changes can be mechanical and testable.

- Build a registry of all exported functions from `#[no_mangle]` and `#[export_name]`, grouped by module and exported symbol.
- Rename non-exported implementation functions away from C names where helpful, for example `XML_Parse` implementation logic can move behind `parse_impl` while `XML_Parse_ffi` remains the ABI shim.
- For currently duplicated pairs such as `XML_ParserCreate` plus `XML_ParserCreate_ffi`, choose one exported shim and one internal implementation path. Keep exported symbols stable.
- Add small conversion helpers:
  - `parser_from_raw(parser: XML_Parser) -> Option<&mut Parser>`.
  - `parser_ref_from_raw(parser: XML_Parser) -> Option<&Parser>`.
  - byte/string range adapters for `(ptr, end)` and `(ptr, len)` inputs.
- Convert exported functions to early-return on null exactly as today, then dispatch to implementation functions that take safe parameters.
- Run `cargo build`, `cargo fmt --all --check`, `cargo check-unsafe2`, and the full C test script.

## Phase 2: Safe Static Data

Purpose: remove global mutable state where the data is actually constant.

- Replace keyword `static mut` arrays in `xmlrole.rs` and `xmltok.rs` with immutable `static` byte arrays or `CStr`/byte-slice constants.
- Replace encoding table `static mut` values in `xmltok.rs` with immutable `static` values once their initialization no longer requires mutation.
- Replace `xmlLen`, `xmlnsLen`, and `implicitContext` with immutable constants or lazily computed safe values.
- For truly mutable process-global settings such as `g_reparseDeferralEnabledDefault` and `g_bytesScanned`, use safe synchronization or atomics:
  - `AtomicBool`/`AtomicU32` where C semantics are plain scalar updates.
  - `OnceLock` for one-time initialized data.
- Remove C2Rust initializer arrays once all static initialization is normal Rust initialization.
- Run the verification gates.

## Phase 3: Safe Tokenizer and Prolog APIs

Purpose: make `xmltok.rs` and `xmlrole.rs` safe internally before changing parser ownership.

- Replace raw pointer ranges in tokenizer implementation functions with slices and cursor structs:
  - input bytes: `&[c_char]` or `&[u8]` behind a small encoding-aware wrapper.
  - output buffers: `&mut [c_char]`, `&mut [u16]`, or a safe writer cursor.
  - next-token outputs: return `(token, next_index)` instead of writing through `*mut *const c_char`.
- Replace internal `unsafe extern "C"` scanner/prolog function pointers with safe Rust function pointer types. Keep ABI shims only for exported `Xml*` functions.
- Convert `PROLOG_STATE.handler` from an unsafe C function pointer to an enum state machine or safe function pointer.
- Convert tokenizer `ENCODING` callback tables to safe methods or safe function pointers. Preserve any C-facing layout only for exported APIs that require it.
- Remove pointer arithmetic by indexing into slices with checked bounds and explicit cursor movement.
- Keep exported tokenizer functions as adapters from raw inputs to safe slices.
- Run the verification gates.

## Phase 4: Parser Ownership and Handles

Purpose: replace the raw `XML_Parser` implementation object with Rust-owned state while preserving the C handle ABI.

- Introduce an internal `Parser` type and make `XML_ParserStruct` an opaque shell for ABI handles if needed.
- Allocate parsers with `Box<Parser>` internally and return `Box::into_raw` cast to the existing `XML_Parser` type at FFI creation functions.
- In `XML_ParserFree`, reconstruct ownership exactly once with `Box::from_raw` and let `Drop` release owned fields.
- Model parent/external entity parser relationships explicitly:
  - Use non-owning identifiers or carefully scoped references where lifetimes are stack-bound.
  - Use `Rc` only if parser/DTD ownership is genuinely shared across external entity parsers and cannot be expressed as parent-owned data.
- Replace nullable parser fields with `Option<T>` where null has semantic meaning.
- Remove `Copy` derives from structs that own resources.
- Run the verification gates.

## Phase 5: Safe Memory Management

Purpose: eliminate `expat_malloc`, `expat_realloc`, `expat_free`, and C allocation from implementation storage.

- Convert parser buffers:
  - `m_buffer` plus pointer triplet fields to `Vec<c_char>` plus index cursors for `buffer_ptr`, `buffer_end`, and `buffer_lim`.
  - `m_dataBuf`/`m_dataBufEnd` to `Vec<XML_Char>` with capacity and current length/index.
  - `m_atts` and `m_nsAtts` to `Vec<ATTRIBUTE>` and `Vec<NS_ATT>`.
- Convert `STRING_POOL` from linked `BLOCK` allocations to a safe pool:
  - store blocks as `Vec<Vec<XML_Char>>` or one arena `Vec<XML_Char>` with stable indices.
  - return pool keys/indices internally instead of raw pointers.
  - only expose raw pointers temporarily when invoking C callbacks, with lifetime tied to the parser call.
- Convert `HASH_TABLE` to `HashMap` or an index-preserving table if collision/order behavior affects tests. Keep the hash salt behavior if externally observable through attack-protection limits.
- Convert `DTD` scaffold arrays to `Vec<CONTENT_SCAFFOLD>` and `Vec<c_int>`.
- Convert tag, binding, and open-entity free lists to owned collections:
  - `Vec<Tag>`/`Vec<Binding>` plus indices for stack links, or `Box` links where identity/stability matters.
  - Remove manual free lists unless profiling shows reuse is still needed; `Vec` capacity reuse is preferred.
- Keep custom user-provided memory suite semantics only at the FFI boundary. If allocation callbacks must remain observable, isolate them behind a safe allocator adapter and document exactly which FFI-returned allocations use it.
- Run the verification gates.

## Phase 6: Strings, Slices, and Callback Data

Purpose: replace raw textual pointers and callback payloads with safe string/slice representations.

- Define safe internal aliases for Expat character data, for example `XmlCharSlice<'a>` and owned `XmlString`.
- Convert null-terminated internal strings to owned `Vec<XML_Char>` or borrowed slices with explicit length.
- Replace `KEY = *const XML_Char` with an owned or interned string key type. Hash tables should own or intern keys so equality and hashing do not depend on pointer identity.
- Replace declaration fields such as doctype names, public IDs, notation names, and current base with `Option<XmlString>` or interned IDs.
- Preserve callback ABI by converting safe string/slice values to stable raw pointers only for the duration required by the callback contract.
- Keep user data (`void *`) raw only in the exported callback ABI wrapper. Internally store it in a transparent `UserData(*mut c_void)` newtype so unsafe-adjacent use is localized and audited.
- Run the verification gates.

## Phase 7: Remove C Library Calls From Implementation

Purpose: stop relying on imported C functions except where the ABI requires it.

- Replace `memset` with `Default`, `fill`, or normal initialization.
- Replace `memcpy`/`memmove` with `copy_from_slice`, `copy_within`, `extend_from_slice`, or `Vec` operations.
- Replace `fprintf(stderr, ...)` debug output with `eprintln!` or a safe debug logging helper.
- Replace `getenv`/`strtoul` with `std::env::var_os` plus Rust parsing, preserving invalid-value behavior.
- Replace `arc4random_buf` with a safe entropy source. Prefer a small dependency only if the standard library cannot provide equivalent behavior on the supported platform.
- Replace `__assert_fail` with `assert!`, `debug_assert!`, or explicit error returns to match existing release behavior as closely as possible.
- Run the verification gates.

## Phase 8: Final Unsafe Elimination

Purpose: make implementation code fully safe and leave unsafety only in audited FFI adapters.

- Remove `unsafe` from every non-exported function once its body and callees are safe.
- Remove raw pointer fields from internal structs. Remaining raw pointers should be limited to:
  - FFI entry point signatures.
  - C callback type definitions required by public API.
  - Narrow conversion helpers at the ABI edge.
- Replace unions such as `TAG` buffer storage with enums that encode the active variant.
- Remove raw pointer casts and int-to-pointer casts from implementation code.
- Re-run `rg` checks for `unsafe`, `*const`, `*mut`, `static mut`, and C imports to confirm remaining instances are only FFI declarations/shims.
- Run `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml`; it must report no implementation unsafe or unsafe-adjacent additions.
- Run the full C test script.

## Suggested Commit Order

1. Add FFI boundary helpers and export registry comments, with no behavior changes.
2. Convert immutable `static mut` data to immutable statics.
3. Safe prolog state machine in `xmlrole.rs`.
4. Safe tokenizer cursor/slice APIs in `xmltok.rs`.
5. Parser `Box` ownership and `Drop` cleanup.
6. Parser buffers and attribute arrays to `Vec`.
7. String pool and DTD scaffold to safe owned storage.
8. Hash tables and interned names to safe collections.
9. Tag, binding, and open-entity lists to owned collections/indices.
10. C library call replacement.
11. Final unsafe removal and formatting normalization.

## Verification Command

Use this full test script after each substantial phase and before declaring the refactor complete:

```sh
set -e
(cd translated_rust && cargo build)
LIBLIBEXPAT_SO=$(readlink -f translated_rust/target/debug/liblibexpat.so)

sh expat/configure --disable-static --enable-shared
touch expat/lib/libtestpat.la
make -C tests -j `nproc` runtests runtests_LDADD="$LIBLIBEXPAT_SO"
ldd tests/runtests | grep "target/debug/liblibexpat.so"
tests/runtests
```

Also run:

```sh
(cd translated_rust && cargo fmt --all --check)
cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml
```

ahomescu:
The plan file for P02:
---
# Safety Refactor Plan for `translated_rust`

## Goal

Refactor the Rust code under `translated_rust` so implementation code is fully safe Rust without changing behavior or the exported ABI.

The end state is:

- FFI entry point signatures are unchanged.
- Only functions with `#[no_mangle]` or `#[export_name]` are treated as FFI entry points.
- FFI entry points perform boundary conversion from C/raw types to safe Rust types, call safe implementation functions, then convert results back to C/raw ABI values.
- All non-entry-point implementation functions are safe Rust functions.
- Implementation code uses safe ownership and borrowing types: `Box`, `Vec`, `Rc`, `Weak`, `&T`, `&mut T`, `&[T]`, `&mut [T]`, `Option`, `Result`, and domain-specific safe structs/enums.
- Implementation code contains no raw pointer fields, raw pointer arguments, raw pointer dereferences, int-to-pointer casts, mutable statics, unions, calls to unsafe functions, or calls through unsafe C function pointers.
- `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml` passes.
- The C test suite still passes.

## Current Inventory

The current tree is a C2Rust-style translation with generated ABI wrappers plus unsafe implementation functions.

From `unsafe_json/translated_rust.json`:

- Total current implementation unsafe findings: `68508`.
- Functions tracked: `13171`.
- FFI entry points tracked by the unsafe inventory: `1896`.
- Non-entry-point implementation functions tracked: `11275`.
- Current FFI-wrapper unsafe boundary findings: `3792`; these are mostly raw ABI signatures plus one call into the unsafe implementation.

Additional source scan:

- Rust files under `translated_rust`: `195`.
- `#[export_name]` occurrences: `1896`.
- `#[no_mangle]` occurrences: `47`.
- `static mut` occurrences: `792`.
- `union` declarations: `74`.
- `extern "C"` import blocks: `184`.
- Raw pointer lines: about `61087`.

Largest implementation unsafe areas by module from the unsafe inventory:

| Unsafe findings | Functions | Module |
| ---: | ---: | --- |
| 5110 | 1017 | `src::cli::cmd` |
| 4182 | 90 | `src::util::hash` |
| 1922 | 288 | `src::libgit2::index` |
| 1917 | 42 | `deps::zlib::trees` |
| 1894 | 275 | `src::libgit2::merge` |
| 1719 | 317 | `src::libgit2::repository` |
| 1472 | 212 | `src::libgit2::submodule` |
| 1463 | 224 | `src::libgit2::checkout` |
| 1349 | 188 | `src::libgit2::iterator` |
| 1307 | 182 | `src::libgit2::remote` |
| 1179 | 171 | `src::libgit2::pack_objects` |
| 1169 | 113 | `src::libgit2::pack` |
| 1133 | 162 | `src::libgit2::refdb_fs` |
| 1126 | 70 | `src::util::str` |
| 1065 | 31 | `deps::zlib::deflate` |

The most important shared causes are raw pointer signatures, raw pointer dereferences, calls into unsafe translated functions, C allocation/string/vector helpers, mutable global state, and C callback/userdata patterns.

## Non-Negotiable ABI Rules

1. Do not change any FFI entry point signature.
   - Keep `unsafe`.
   - Keep `extern "C"`.
   - Keep argument and return types exactly as they are.
   - Keep `#[no_mangle]` and `#[export_name]` attributes and exported names.

2. Do not classify a function as an FFI entry point only because it is `unsafe extern "C"`.
   - A function is an entry point only if it has `#[no_mangle]` or `#[export_name]`.
   - All other functions are implementation functions and must become safe.

3. Preserve public-layout ABI types.
   - Before changing a `#[repr(C)]` struct layout, check whether it is visible by value in C headers/tests or returned/accepted by value.
   - For ABI-visible layout structs, keep a C wire type and convert to a safe internal type.
   - For structs that only appear behind pointers and are intended to be opaque, freely replace fields with safe Rust types.

4. Keep unsafe boundary code minimal.
   - The only remaining unsafe operations should be in FFI entry points or tiny boundary conversion helpers called only by those entry points.
   - Do not move implementation logic into FFI entry points.

## Refactoring Strategy

Use a two-layer design throughout:

1. `ffi` or `wire` layer:
   - Keeps exported function signatures and public C structs.
   - Validates null pointers, lengths, aliasing assumptions, and output pointers.
   - Converts raw inputs to safe Rust values.
   - Converts safe outputs back to ABI values.
   - Owns all unavoidable C callback invocation and raw allocation handoff.

2. Safe implementation layer:
   - Contains the real behavior.
   - Uses references, slices, owned containers, `Rc`/`Weak`, and domain-specific enums.
   - Returns `Result<T, GitError>` or existing libgit2-style integer codes via safe wrappers.
   - Has no raw pointer types in signatures or fields.
   - Has no `unsafe fn` and no `unsafe` blocks.

Apply this module by module, but start with shared utility types. Most libgit2 modules depend on these helpers, so making them safe first removes unsafe calls at scale.

## Phase 0: Baseline and Guardrails

1. Record the baseline.
   - Run the existing C test script once on the unrefactored tree.
   - Run `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml` and save the initial report.
   - Keep `unsafe_json/translated_rust.json` as the initial inventory until a fresh report is produced.

2. Add local tracking commands.
   - Keep a small notes file or script with:
     - count of FFI entry points,
     - count of implementation unsafe findings,
     - count of raw pointer fields/signatures,
     - modules completed.

3. Establish the per-change gate.
   - After every coherent module conversion:
     - `cargo build --manifest-path translated_rust/Cargo.toml`
     - `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml`
   - After each phase:
     - run the full C test script from this plan's final section.

## Phase 1: Build the Boundary Conversion Toolkit

Create safe helpers used by FFI entry points. These helpers should be the only code that turns raw C inputs into safe Rust values.

1. Pointer conversion helpers.
   - `nonnull_ref<T>(*const T) -> Result<&T, GitError>`
   - `nonnull_mut<T>(*mut T) -> Result<&mut T, GitError>`
   - `nullable_ref<T>(*const T) -> Option<&T>`
   - `nullable_mut<T>(*mut T) -> Option<&mut T>`
   - `slice_from_raw_parts<T>(*const T, len: usize) -> Result<&[T], GitError>`
   - `slice_from_raw_parts_mut<T>(*mut T, len: usize) -> Result<&mut [T], GitError>`

2. C string helpers.
   - Convert `*const c_char` to a safe borrowed C string view at the boundary.
   - Convert validated text to `&str` only where existing behavior requires valid UTF-8.
   - Preserve byte-oriented semantics for git paths and object data by using byte slices, not Rust strings.

3. Output pointer helpers.
   - Centralize writing `out` parameters.
   - Preserve existing null-output behavior exactly.
   - Make successful ownership transfer explicit with `Box::into_raw` only at the FFI boundary.

4. Error conversion.
   - Introduce a safe `GitError`/`GitResult<T>` representation.
   - Keep libgit2 integer error codes at the FFI boundary.
   - Replace imported `git_error_set*` calls inside implementation with safe Rust error construction and boundary-level error publication.

5. Callback adapters.
   - Convert C callback function pointer plus `payload: *mut c_void` into a safe closure or trait object in the FFI entry point.
   - Internal APIs take `Fn`, `FnMut`, or domain callback traits.
   - C callback invocation remains boundary code; implementation code only calls the safe closure/trait.

Acceptance gate:

- New helper modules must not introduce unsafe findings in implementation code according to `cargo check-unsafe2`.
- No exported signatures change.

## Phase 2: Replace Core Allocation and Ownership

The generated code currently routes most allocations through `git__allocator`, `git__malloc`, `git__calloc`, `git__reallocarray`, and `git__free`.

1. Replace internal allocations.
   - Use `Box<T>` for single owned objects.
   - Use `Vec<T>` or `Box<[T]>` for arrays.
   - Use `String`, `CString`, or `Vec<u8>` for owned text depending on byte/UTF-8 requirements.
   - Use `Rc<T>` and `Weak<T>` for shared graph/cache ownership.

2. Preserve allocator ABI functions.
   - Keep exported allocator setup functions and signatures.
   - At the boundary, support custom C allocator callbacks where tests require them.
   - Internally, do not call allocator function pointers directly. Route allocation through safe Rust containers and map allocation failure to existing error codes.

3. Replace manual zeroing/copying.
   - `memset` zeroing -> `Default`, `fill`, or construction with zero values.
   - `memcpy`/`memmove` -> `copy_from_slice`, `copy_within`, `Vec::extend_from_slice`, `clone_from_slice`, or assignment.
   - `memcmp` -> slice comparison.

4. Replace ownership-transfer conventions.
   - Functions that currently return `*mut T` internally should return `Box<T>`, `Rc<T>`, or a domain handle.
   - FFI wrappers convert `Box<T>` to raw only when returning ownership to C.
   - Free entry points reconstruct ownership and drop at the boundary.

Acceptance gate:

- `src::util::alloc` has no implementation unsafe.
- Internal callers no longer import or call `git__malloc`, `git__calloc`, `git__reallocarray`, `git__free`, or allocator function pointers.
- Allocation failure behavior remains covered by existing tests.

## Phase 3: Convert Shared Utility Types

These modules should be made safe before the larger libgit2 modules because they remove repeated unsafe calls.

1. `git_str` / `git_buf` in `src::util::str`.
   - Implement a safe internal buffer backed by `Vec<u8>`.
   - Preserve the trailing NUL invariant where C callers observe `ptr`.
   - Replace `git_str__initstr` and `git_str__oom` mutable sentinels with immutable constants or enum state.
   - Use safe methods for grow, put, splice, trim, encode/decode, detach, and attach.
   - Keep ABI wrappers for C-visible `git_str`/`git_buf` operations.

2. `git_vector` in `src::util::vector`.
   - Replace `contents: *mut *mut c_void` and manual capacity with `Vec<T>` or a domain-safe erased handle type.
   - Convert comparator callbacks to safe comparator traits/closures at boundaries.
   - Use safe sorting, binary search, insertion, removal, and range operations.

3. `pqueue`, `sortedcache`, `strlist`, and `oidarray`.
   - Use `Vec`, `BinaryHeap` where ordering matches behavior, and safe iterator APIs.
   - For C arrays returned to callers, allocate owned Rust arrays and expose raw views only in FFI wrappers/free functions.

4. Path and byte utilities.
   - Replace C string scanning with slice iteration.
   - Preserve byte-for-byte path behavior; do not impose UTF-8 except where the original did.
   - Use `std::path` only where it does not alter libgit2 path semantics.

Acceptance gate:

- `src::util::{str,vector,pqueue,sortedcache,strlist,oidarray,fs_path,futils,util}` implementation functions have no unsafe findings.
- Major downstream modules compile against safe utility APIs.

## Phase 4: Remove Mutable Statics and Unions

1. Mutable statics.
   - Convert immutable lookup tables to `static` or `const`.
   - Convert lazily initialized globals to `OnceLock`, `LazyLock`, `Mutex`, `RwLock`, or `Atomic*` depending on mutation and thread-safety needs.
   - Replace global config maps with immutable slices.
   - Replace global allocator state with a safe synchronized value if it must remain configurable.

2. Unions.
   - Replace C2Rust generated unions with enums when the active field is semantic.
   - Replace punning unions that store two integer interpretations with named safe fields or methods.
   - For zlib tree data (`freq/code`, `dad/len`), use a safe struct with explicit fields if both values are used at different algorithm phases.

3. Constants.
   - Replace `transmute`-built byte constants with direct byte arrays or safe `c"..."`-style equivalents where the compiler/toolchain supports them.

Acceptance gate:

- No implementation use of `static mut`.
- No implementation use of union fields.
- No new global mutation without synchronization.

## Phase 5: Replace FFI Imports with Rust Calls or Safe Standard APIs

Generated modules import functions from each other through `extern "C"` blocks. These are implementation calls, not real foreign ABI needs.

1. Internal Rust-to-Rust calls.
   - Replace `extern "C"` imports of functions implemented in this crate with direct Rust module imports.
   - Call safe implementation functions, not exported ABI wrappers.
   - Delete redundant import declarations as modules are converted.

2. libc memory/string functions.
   - Replace with safe slice and standard-library operations.

3. filesystem/process/time calls.
   - Replace `open`, `creat`, `read`, `write`, `close`, `rename`, `unlink`, `stat`, `chmod`, `getcwd`, `realpath`, and environment helpers with `std::fs`, `std::io`, `std::env`, and `std::path` where behavior matches.
   - Keep C ABI wrappers such as `p_open`, `p_read`, and `p_write`, but make them boundary adapters over safe implementations.

4. mmap.
   - Prefer safe file-backed buffers if tests do not require true memory mapping behavior.
   - If memory-map semantics must remain externally visible, isolate the OS mapping in FFI/boundary code and expose a safe internal mapped-buffer abstraction.

Acceptance gate:

- Implementation functions no longer call FFI imports.
- Remaining `extern "C"` blocks are only for true ABI boundary requirements.

## Phase 6: Convert zlib

The zlib subtree is algorithmically contained and should be converted as a unit after buffer and allocation utilities are safe.

1. State ownership.
   - Replace heap-allocated `deflate_state` and `inflate_state` internals with `Box<DeflateState>` and `Box<InflateState>`.
   - Keep `z_stream_s` ABI fields unchanged at FFI entry points.
   - Convert `state` raw pointers at the boundary and store safe state internally.

2. Input/output windows.
   - Replace `next_in`, `next_out`, `window`, `pending_buf`, `prev`, `head`, and `sym_buf` pointer arithmetic with slices, `Vec`, and cursor indices.
   - Use checked indexing and explicit bounds decisions that match the original C behavior.

3. Tables.
   - Convert `static mut` tables in `crc32`, `trees`, `inflate`, and `inftrees` to immutable arrays or `OnceLock`.
   - Replace zlib `ct_data` unions with explicit safe fields.

4. Function-pointer state machines.
   - Replace `compress_func` unsafe C function pointers with Rust enum dispatch or safe function items.

Acceptance gate:

- `deps::zlib::*` implementation functions have no unsafe findings.
- zlib exported ABI tests continue to pass through the existing libgit2 test suite.

## Phase 7: Convert xdiff

The xdiff subtree is pointer-heavy but mostly self-contained.

1. Memory files and records.
   - Convert `mmfile`, records, hash tables, and change-script structures to safe slices, vectors, and owned structs.
   - Replace linked-list raw pointers with `Vec` indices or `Option<Box<Node>>` where lists are truly needed.

2. Callback output.
   - Convert emit callbacks at FFI entry points to safe closures.
   - Internal diff/merge code calls safe emit traits.

3. Algorithms.
   - Convert pointer walking to index-based loops over slices.
   - Keep exact output formatting and merge conflict behavior.

Acceptance gate:

- `deps::xdiff::*` implementation functions have no unsafe findings.
- Existing diff, patch, merge, and xdiff-dependent C tests pass.

## Phase 8: Convert libgit2 Domain Objects

Convert object families in dependency order. For each family, split ABI wire types from safe internal domain types when necessary.

1. Repository and shared ownership.
   - Convert `git_repository` internals to a safe owned struct.
   - Use `Rc<RepositoryInner>` for shared ownership and `Weak` for back references from objects/caches.
   - Replace raw back-pointers with safe references, IDs, or `Weak` handles.
   - Convert repository-owned caches, config, odb, index, worktree, and submodule state to safe containers.

2. Object database and object model.
   - Convert `git_odb`, backends, loose/pack/mempack, `git_object`, `git_commit`, `git_tree`, `git_blob`, and `git_tag`.
   - Use `Box` for unique ownership, `Rc` for cached objects, and safe byte buffers for object contents.
   - Replace parser functions that take raw data pointers with slice-based parsers.

3. Index, tree, iterator, checkout.
   - Convert arrays of entries to `Vec<Entry>`.
   - Convert iterator vtables from function pointers over raw objects to Rust traits.
   - Convert checkout/apply path workflows to safe path and buffer operations.

4. Config, refs, refdb, reflog.
   - Replace backend vtables with safe traits.
   - Convert string ownership to safe buffers.
   - Replace raw iterator ownership with boxed trait objects or concrete iterator structs.

5. Diff, patch, merge, blame, status.
   - Build on safe xdiff and safe vectors.
   - Replace callback/userdata patterns with safe callback traits created at the FFI boundary.
   - Use safe owned lists for deltas, hunks, lines, conflicts, and similarity candidates.

6. Remote, transport, fetch, push, submodule, worktree.
   - Convert nested callback structures at entry points.
   - Use safe owned option/config structs internally.
   - Preserve externally visible callback order, error propagation, and cancellation behavior.

Acceptance gate for each object family:

- No implementation unsafe findings in the converted module family.
- No exported function signature changes.
- Existing C tests for that family pass.

## Phase 9: Convert CLI Code

The CLI is not an FFI surface, so it can be refactored more freely.

1. Replace C stdio imports with Rust `std::io`.
2. Replace command option arrays and parser pointer state with safe structs and slices.
3. Replace raw argv handling with `Vec<String>`/`OsString` or byte-aware equivalents where needed.
4. Convert command callbacks to safe function pointers or enum dispatch.

Acceptance gate:

- `git2_cli` builds.
- `src::cli::*` implementation functions have no unsafe findings.

## Phase 10: Final Wrapper Pass

After implementation functions are safe, update every exported function to be a thin wrapper.

For each `#[export_name]` or `#[no_mangle]` function:

1. Keep the exact current signature.
2. Validate pointers and lengths according to existing behavior.
3. Convert inputs to safe references, slices, owned values, or callback adapters.
4. Call the safe implementation function.
5. Convert the result back to ABI form.
6. Publish errors through existing libgit2 error APIs.
7. Keep any unsafe block small and local to the conversion.

For every non-exported function:

1. Remove `unsafe`.
2. Remove `extern "C"` unless there is a concrete ABI reason.
3. Remove raw pointer arguments and returns.
4. Remove raw pointer fields from structs.

Acceptance gate:

- `cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml` passes with no implementation unsafe or unsafe-adjacent findings.

## Behavior Preservation Practices

1. Use characterization tests before risky rewrites.
   - For small pure functions such as varint, hash formatting, base encoders, path matching, and wildcard matching, add focused Rust tests around current behavior before converting.
   - For libgit2 behavior, rely primarily on the C test suite.

2. Convert one module family at a time.
   - Do not combine semantic cleanup with safety conversion.
   - Keep names and error codes close to the original until tests are green.

3. Preserve byte semantics.
   - Git paths, object contents, diffs, and config data are byte-oriented in many places.
   - Avoid converting to `str` unless the original behavior required valid text.

4. Preserve allocation and free contracts.
   - If C receives ownership from an exported function, C must still free it through the expected exported free function.
   - Internally, use Rust ownership; only leak/restore `Box` ownership at FFI boundaries.

5. Preserve callback behavior.
   - Maintain callback invocation order.
   - Maintain cancellation/error propagation.
   - Keep callback payload lifetime tied to the FFI call unless existing API documents longer retention.

6. Preserve null handling.
   - Match existing behavior for nullable inputs, optional outputs, and null-as-default option structs.
   - Document each FFI wrapper's null policy while converting it.

## Suggested Module Order

Use this order to maximize reuse of safe foundations:

1. Boundary helpers and error/result model.
2. `src::util::alloc`.
3. `src::util::str`, `buf`, and related string/path helpers.
4. `src::util::vector`, `pqueue`, `sortedcache`, `strlist`, `oidarray`.
5. Immutable tables, globals, and union replacements.
6. `deps::zlib`.
7. `deps::xdiff`.
8. `src::libgit2::repository`, `object`, `object_api`, `odb`, `odb_*`.
9. `src::libgit2::config`, `refs`, `refdb`, `reflog`.
10. `src::libgit2::index`, `tree`, `iterator`, `checkout`.
11. `src::libgit2::diff`, `patch`, `merge`, `apply`, `status`, `blame`.
12. `src::libgit2::remote`, `transport`, `fetch`, `push`, `submodule`, `worktree`.
13. Remaining libgit2 modules.
14. CLI modules.
15. Final wrapper-only ABI pass.

## Final Verification

Run the required build and C tests:

```sh
set -e
cd .
cargo build --manifest-path translated_rust/Cargo.toml
RUST_SO="$PWD/translated_rust/target/debug/liblibgit2package.so"
cmake -S tests -B tests/build --preset test -DLIB_LINK_PATH="$RUST_SO"
cmake --build tests/build
./tests/build/tests
```

Run the unsafe checker:

```sh
cargo check-unsafe2 --manifest-path translated_rust/Cargo.toml
```

The refactor is complete only when both commands pass and no implementation function contains unsafe or unsafe-adjacent code.

ahomescu:
Phase 6 of the libgit2 migration is "Convert zlib" which was a pure but lucky coincidence.

crisp/sandbox/docker.py
```diff
@@ -83,6 +83,8 @@ def commit_dir(self, rel_path, ignore_spec: PathSpec | None = None):
         dest_prefix = os.path.dirname(rel_path)
         with tarfile.open(fileobj=tar_io, mode='r') as t:
             while (info := t.next()) is not None:
+                if ignore_spec is not None and ignore_spec.match_file(info.name):
```
ahomescu, line 91: I split this into a separate PR, it should be useful by itself https://github.com/GaloisInc/Tractor-Crisp/pull/153.

crisp/workflow.py
```diff
@@ -145,7 +145,7 @@
 '''
 
 AGENT_SAFETY_PROMPT = '''
-Continue the plan from `SAFETY_PLAN.md`.
+Continue the plan from `SAFETY_PLAN.md`, preferably using sub-agents (without full history) to make progress in parallel.
```
thedataking: I'd skip this for now. The [OpenAI docs](https://learn.chatgpt.com/docs/agent-configuration/subagents?surface=app#why-subagent-workflows-help) does not recommend subagents for code editing:

> As a starting point, use parallel agents for read-heavy tasks such as exploration, tests, triage, and summarization. Be more careful with parallel write-heavy workflows, because agents editing code at once can create conflicts and increase coordination overhead.

thedataking:
I'm experimenting with a version of this which is more prescriptive about subagents to spawn, what the subagents should do (this needs a lot more refining but the direction should be an improvement) in #160.  

`AGENT_PLAN_PROMPT` now also specifies how the plan will be used and what overall format it should have. Hopefully, this will lead to more deterministic planning and lifting trajectories.

crisp/workflow.py
ahomescu: Removing sub-agents from both planning and execution.

ahomescu:
@thedataking I removed all the prompt changes except the planning split, then we could land this PR and stack #160 on top.

crisp/workflow.py
```diff
@@ -1289,6 +1292,41 @@ def do_safety_step_llm(
 
         return self.do_validate_and_repair(n_code, n_new_code, n_test_code)
 
+    @step
+    def do_safety_plan_agent(
+        self,
+        n_code: TreeNode,
+        n_test_code: TreeNode,
+    ) -> TreeNode:
```
thedataking: `run_rewrite` returns `tuple[TreeNode, TreeNode]`.

crisp/agent.py
```diff
@@ -24,15 +24,16 @@ def _snapshot_to_family_alias(model: str) -> str:
     m = _SNAPSHOT_SUFFIX.match(model)
     return m.group("alias") if m else model
 
-def _codex_command(cfg: Config, subcmd: str, args: list[str], codex_login: bool = False) -> list[str]:
+def _codex_command(cfg: Config, subcmd: str, args: list[str],
+                   default_model: str, codex_login: bool = False) -> list[str]:
```
thedataking: why not `model`, that would match the name of the parameter you're passing to `codex`?

I find this  confusing because we almost always want to say "use this specific model" and never use the override in the environment. 

In fact, the whole thing would become simpler if we got rid of the `CRISP_` environment variables and relied entirely on config for model selection. We would still need to pass API keys in the environment but that could use the names already read by the API endpoints (`OPENAI_API_KEY`, `GEMINI_API_KEY`, etc.) which would make it easy to use models from different families in different steps.

crisp/agent.py
```diff
@@ -98,6 +99,7 @@ def run_rewrite(
     cfg: Config,
     mvir: MVIR,
     prompt: str,
+    default_model: str,
```
thedataking: `model`...

crisp/workflow.py
```diff
@@ -1197,7 +1197,8 @@ def agent_safety(
         )
         if prompt_suffix is not None:
             prompt = f'{prompt}\n\n{prompt_suffix}'
-        return agent.run_rewrite(cfg, mvir, prompt, n_code,
+        default_model = self.cfg.models.agent_loop
+        return agent.run_rewrite(cfg, mvir, prompt, default_model, n_code,
```
thedataking: `agent.run_rewrite(cfg, mvir, prompt, cfg.models.agent_loop, n_code, ...`  would be so much clearer.

same comment around lines 1320-1321.

crisp/workflow.py
```diff
@@ -1289,6 +1292,41 @@ def do_safety_step_llm(
...
+        if cfg.test_command is not None:
+            after_refactoring_instruction = AGENT_AFTER_REFACTORING_RUN_TESTS \
+                    .format(test_cmd = cfg.test_command)
+        else:
+            after_refactoring_instruction = AGENT_AFTER_REFACTORING_BUILD
+
+        extra_code = [
+            self.find_unsafe2_json(n_code),
+        ]
+        if n_test_code is not None:
```
thedataking, line 1312: Not sure we should do this if we're in `agent_sim_no_tests` mode which withholds tests from the agent.

`agent_safety` has a `provide_test_cmd` flag; maybe we should do something similar here?

crisp/workflow.py
thedataking: I think you need to say that these rules must be copied into the `SAFETY_PLAN.md` verbatim like we do in #160. The suffix we might add in `__main__.py` says "NOT from FFI entry points" which wouldn't make sense unless these instructions are in the `SAFETY_PLAN.md`.

crisp/workflow.py
```diff
@@ -1289,6 +1292,42 @@ def do_safety_step_llm(
...
+            extra_code.append(n_test_code)
+
+        prompt = AGENT_PLAN_PROMPT.format(
+            cargo_dir_path = cargo_dir,
+            after_refactoring_instruction = after_refactoring_instruction,
+        )
+        default_model = self.cfg.models.agent_plan
+        return agent.run_rewrite(cfg, mvir, prompt, default_model, n_code,
```
thedataking: we didn't change the code, so when `run_rewrite` constructs the `CodexAgentOpNode`, `old_code` and `new_code` will be the same node. 

then a subsequent safety step may also produce a no-op where `old_code` equals `new_code`

on resume, `prior_agent_plans` gets called. I think you might hit the case where it finds multiple matches and raises `CrispError` (`__main__.py:294`). I don't think that logic was ever ideal, maybe we should just pick the newest node instead of giving up?

thedataking (changes requested):
nit: you're splitting `agent` into `agent_loop` and `agent_plan` but README.md example still says `models.agent`

crisp/agent.py
ahomescu: Getting rid of the environment variables is out of scope for this PR, I'll just change the variable name.

crisp/workflow.py
ahomescu: My thinking was the opposite: we still want the separate planning even without tests. That said, I'm not sure we should have a "no_tests" mode at all.

crisp/workflow.py
ahomescu: I fixed that in one of the commits, it ignores nodes where `old_code == new_code`. Is there another problem?

ahomescu: See https://github.com/GaloisInc/Tractor-Crisp/pull/152/changes/518c858027749ccc66add584d342129e04e398a6.

crisp/workflow.py
thedataking: > we still want the separate planning even without tests

yes, definitely. I wasn't being clear: I'm saying that we shouldn't pass any tests to the planner if we're in `agent_sim_no_tests` mode. If we do so, the safety plan might reference testing files or identifiers which then confuses the subsequent safety rewrite step that can't see the tests.

Since there is a no-test mode, we should make sure this PR doesn't introduce regressions.

crisp/workflow.py
thedataking: > See https://github.com/GaloisInc/Tractor-Crisp/commit/518c858027749ccc66add584d342129e04e398a6.

That fixes the history in `render`. I'm saying we can hit the `raise CrispError` case here:

```Python
def prior_agent_plans(mvir, n_code) -> TreeNode | None:
    # Look up the node which produced `n_code` and return the planning files for that step.
    # This lets a resumed safety-loop pick up the previous `SAFETY_PLAN.md` if it exists.
    matches = [
        ie for ie in mvir.index(n_code.node_id())
        if ie.kind == CodexAgentOpNode.KIND and ie.key == 'new_code'
    ]

    match matches:
        case [ie]:
            op_node = mvir.node(ie.node_id)
            return mvir.node(op_node.planning_files)
        case []:
            return None
        case _:
            raise CrispError(
                f'multiple Codex agent ops produced code node {n_code.node_id()}: '
                + ', '.join(str(ie.node_id) for ie in matches)
            )
```

WDYT about doing this instead:
```Python
def prior_agent_plans(mvir, n_code) -> TreeNode | None:
    # Look up the node which produced `n_code` and return the planning files for that step.
    # This lets a resumed safety-loop pick up the previous `SAFETY_PLAN.md` if it exists.
    matches = set(
        ie.node_id for ie in mvir.index(n_code.node_id())
        if ie.kind == CodexAgentOpNode.KIND and ie.key == 'new_code'
    )
    if not matches:
        return None

    # Multiple ops can produce the same tree (planning step, no-op safety
    # steps); take the most recent one from the `op_history` reflog.
    op_node_id = None
    if mvir.has_tag('op_history'):
        for entry in mvir.tag_reflog('op_history'):
            if entry.node_id in matches:
                op_node_id = entry.node_id
    if op_node_id is None:
        op_node_id = matches.pop()

    op_node = mvir.node(op_node_id)
    return mvir.node(op_node.planning_files)
```

crisp/workflow.py
ahomescu: Looking at the call sites, it seems that `n_test_code` is a misnomer here anyway, because what actually gets passed in is the full C code. We can discuss whether that should be passed to the planner or not, but that's an orthogonal discussion to testing. See a later comment in `crisp/workflow.py`:
```
        # Don't provide the test code, so the agent can't
        # accidentally find the tests.  Note this has the side
        # effect of not providing the original C code, since we
        # don't currently distinguish test code from the rest of
        # the C code.
```

crisp/workflow.py
ahomescu: > then a subsequent safety step may also produce a no-op where `old_code` equals `new_code`

This shouldn't happen right now, it would mean that the agent ran for one iteration and produced absolutely nothing.

ahomescu: Update from offline discussion: we concluded that this could happen, but it was a preexisting issue that predates this PR.

thedataking: from offline discussion: we're going to defer this issue. Issue not introduced in this PR.

Note: we haven't hit this case where we end up raising `CrispError` but we could with a harder project/dumber LLM but we can cross that bridge when we get there.

crisp/workflow.py
ahomescu: I added a TODO and comment.

thedataking (approved):

crisp/workflow.py
thedataking: It appears https://github.com/GaloisInc/Tractor-Crisp/pull/163 fixes this.

PR #153, opened 2026-07-07
Check the ignore_spec before the file type

ahomescu:
Fix an ordering issue: if the sandbox contains a file of an unknown type (e.g. a symlink) that's also on the ignore list, check the list first and discard the file. Otherwise the commit step errors out on that file.


spernsteiner (approved):
Looks like `sandbox.bwrap` / `crisp.work_dir` already does things in the right order, so no need to change that one

Issue #146, opened 2026-06-12
Crisp fails on json-c due to brittle rustfmt error handling


Crocodoctopus:
When running Crisp, rustfmt exits with non-zero during `split_ffi_op` while formatting transpiled Rust output produced by c2rust on commit 703aa4c7b0. This is brittle, as the code formats correctly for all other lines, except for these two.

The failing Crisp output:
```
warning: unused config key `unstable.sparse-registry` in `/home/akourne/repos/c2rust/.cargo/config.toml`
error[internal]: left behind trailing whitespace
    --> /home/akourne/repos/c2rust/json-c-transpile/src/json_tokener.rs:1563:1563:1
     |
1563 |
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |

error[internal]: left behind trailing whitespace
    --> /home/akourne/repos/c2rust/json-c-transpile/src/json_tokener.rs:1627:1627:1
     |
1627 |
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
```

The failing Rust code in `json_tokener.rs:1563`:
```rust
                                                            crate::src::printbuf::printbuf_memappend(
                                                                
                                                                (*tok).pb as *mut crate::src::printbuf::printbuf,
                                                                &raw mut utf8_replacement_char
                                                                    as *mut ::core::ffi::c_uchar
                                                                    as *mut ::core::ffi::c_char,
                                                                3 as ::core::ffi::c_int,
                                                            );
```

`json_tokener.rs:1627` is basically the same.

I suspect the long leading indentation is the cause, as increasing the max width fmt allows resolve this issue, though I doubt that is the only way rustfmt can give this class of failure.

I don't think Crisp should assert rustfmt must be return 0, but I'm not sure if we should just relax these cases to a warning.

PR #154, opened 2026-07-08
Misc fixes for P02

spernsteiner:
* Bump C2Rust submodule to get the reorganize-definitions fix in immunant/c2rust#1859
* Bump nightly toolchain for tools (2026-05-11 -> 2026-06-17) to fix a `rustc_public` (Stable MIR) panic on associated type impls
* Add filenames to `find-unsafe2` outputs
* Adjust random targeting to focus on one file at a time
* Various minor changes to the agent prompt

### DIFF
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -521,6 +521,11 @@ class PickTarget:
 
     def current_target_goal(self, w, n_code):
         # Keep going with the current goal if possible.
+
+        # Optimistically take fuel from both the target and file to continue
+        # with the current goal.  If the target is out of fuel, we change
+        # targets and `give()` the target more fuel below; if the file is out
+        # of fuel, we changes files and targets and `give()` both more fuel.
         target_had_fuel = self.fuel_target.try_use()
         file_had_fuel = self.fuel_file.try_use()
 
@@ -565,8 +570,13 @@ class PickTarget:
         # Found no goal in any file, but `count_unsafe2` still reported some
         # unsafe code.  Tell the agent to check the entire codebase for
         # leftover unsafety.
+        self.target_goal = AgentTargetOther()
+        self.fuel_file.give()
+        self.fuel_file.use()
+        self.fuel_target.give()
+        self.fuel_target.use()
         print('PickTarget: no valid targets')
-        return AgentTargetOther()
+        return self.target_goal
 
 
 def pick_file_and_list_targets(w, n_code):
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -1,5 +1,6 @@
 import argparse
 import ast
+from collections import defaultdict
 from dataclasses import dataclass
 import glob
 import json
@@ -27,7 +28,7 @@ from .mvir import MVIR, NodeId, FileNode, TreeNode, LlmOpNode, \
 from .sandbox import run_sandbox
 from .work_dir import lock_work_dir, set_keep_work_dir
 from .workflow import (
-    Workflow, OutOfFuelError, AgentTargetField, AgentTargetFunction,
+    Workflow, FuelCounter, OutOfFuelError, AgentTargetField, AgentTargetFunction,
     AgentTargetOther,
 )
 
@@ -322,6 +323,10 @@ class FuelLimits:
     # the outer loop will pick a new target immediately instead.
     safety_tries_per_target: int
 
+    # Give the agent this many iterations within each file before swiching to a
+    # new file.
+    safety_tries_per_file: int
+
 def total_code_size(mvir, n_code):
     total = 0
     for name, file in n_code.files.items():
@@ -341,6 +346,7 @@ def get_fuel_limits(mvir, n_code):
             safety_tries = 8,
             max_consecutive_failures = 3,
             safety_tries_per_target = 2,
+            safety_tries_per_file = 10,
         )
     elif size < 20000:
         # P01
@@ -348,6 +354,7 @@ def get_fuel_limits(mvir, n_code):
             safety_tries = 45,
             max_consecutive_failures = 5,
             safety_tries_per_target = 3,
+            safety_tries_per_file = 20,
         )
     else:
         # P02 - run forever
@@ -355,6 +362,7 @@ def get_fuel_limits(mvir, n_code):
             safety_tries = 9999,
             max_consecutive_failures = 9999,
             safety_tries_per_target = 5,
+            safety_tries_per_file = 50,
         )
     print(f'code size = {size}')
     print(f'default limits = {defaults!r}')
@@ -369,6 +377,9 @@ def get_fuel_limits(mvir, n_code):
         safety_tries_per_target = int(
             os.environ.get('LLM_SAFETY_TRIES_PER_TARGET',
                 defaults.safety_tries_per_target)),
+        safety_tries_per_file = int(
+            os.environ.get('LLM_SAFETY_TRIES_PER_FILE',
+                defaults.safety_tries_per_file)),
     )
 
 def safety_loop_common(args, cfg, mvir, w, n_code, n_c_code):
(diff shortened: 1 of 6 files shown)


thedataking:
@spernsteiner the c2rust updates needed are in `master` now; please update `c2rust` submodule accordingly.

spernsteiner:
Rebased over https://github.com/GaloisInc/Tractor-Crisp/pull/151, so we no longer need to prompt the agent to run `cargo clean` before `cargo check-unsafe2`.

@thedataking, I added d79c5993829483d73b2a921730313e4be9e0f4fd to fix the find-unsafe2 errors related to exposed provenance that you mentioned previously.  The issue there is that the exposed provenance APIs were renamed in 2024, in between the 2023 nightly that c2rust-transpile targets and the 2026 nightly used by find-unsafe2.  The fix applied here is to upgrade the transpiler output to the 2026 nightly by patching `rust-toolchain.toml` and find-and-replacing all uses of the exposed provenance APIs.

thedataking:
> @thedataking, I added https://github.com/GaloisInc/Tractor-Crisp/commit/d79c5993829483d73b2a921730313e4be9e0f4fd to fix the find-unsafe2 errors related to exposed provenance that you mentioned previously.

Sounds good to me.

non-blocking nits:

 - `deps/c2rust` is bumped to https://github.com/immunant/c2rust/pull/1900. 
it would be good to also include https://github.com/immunant/c2rust/pull/1902 since it is libgit2 related

no action required:

I came across https://crates.io/crates/sptr which polyfills the old APIs. I think the rewriting commit you added is the right approach; not suggesting we use `sptr`. Just sharing since it gives us a way to have c2rust output code that can be refactored and then compile on newer toolchains.

crisp/workflow.py
```diff
@@ -999,6 +999,14 @@ def split_ffi_op(self, n_tree: TreeNode) -> SplitFfiOpNode:
                 exit_code, logs2 = sb.run([
                     'cargo', 'fmt', '--manifest-path',
                     sb.join(rust_path_rel, 'Cargo.toml')])
+
+                # `cargo fmt` sometimes fails with an internal error about
+                # being unable to remove trailing whitespace.  We log the error
+                # in that case and then suppress it.
+                if exit_code != 0:
```
thedataking, line 1096: nit This suppresses all errors; could look for:

```
error[internal]: left behind trailing whitespace
```

in the output. 

Once this happens, it seems to repeat over and over, so we might want to file an issue to fix the underlying cause and reference that issue here?

crisp/workflow.py
```diff
@@ -788,6 +792,65 @@ def patch_build_rs_op(self, code: TreeNode, libs: list[str]) -> EditOpNode:
...
+    def patch_upgrade_toolchain_op(self, code: TreeNode) -> EditOpNode:
+        cfg, mvir = self.cfg, self.mvir
+
+        new_files = {}
+        for path, file_id in code.files.items():
+            if path.endswith('.rs'):
+                file = mvir.node(file_id)
+                old_src = file.body_str()
+                new_src = (old_src
+                    .replace('ptr::from_exposed_addr', 'ptr::with_exposed_provenance')
```
thedataking, line 847: This correctly handles [`from_exposed_addr_mut`](https://rust-for-linux.github.io/docs/core/ptr/fn.from_exposed_addr_mut.html#). Worth a comment?

crisp/workflow.py
```diff
@@ -788,6 +792,65 @@ def patch_build_rs_op(self, code: TreeNode, libs: list[str]) -> EditOpNode:
...
+                    .replace('.expose_addr()', '.expose_provenance()'))
```
thedataking: I'm seeing a lot of messages like this in the log:

```
[03:43:13    32] warning: the feature `raw_ref_op` has been stable since 1.82.0 and no longer requires an attribute to enable
[03:43:13    32]  --> lib.rs:9:12
[03:43:13    32]   |
[03:43:13    32] 9 | #![feature(raw_ref_op)]
```

Should we also remove that to shorten the log and output the LLM has process?

crisp/workflow.py
```diff
@@ -215,15 +215,34 @@ class AgentTargetOther(AgentTarget):
...
 
-    def give(self, amount):
+    def try_use(self) -> bool:
+        if self.fuel == 0:
+            return False
+        else:
+            self.fuel -= 1
+            return True
+
+    def is_empty(self) -> bool:
```
thedataking, line 241: nit: appears unused.

crisp/__main__.py
```diff
@@ -491,15 +495,80 @@ def safety_loop_common(args, cfg, mvir, w, n_code, n_c_code):
...
+            self.fuel_target.use()
+            print(f'PickTarget: new file {self.current_file}, target {self.target_goal}')
+            print(f'  target fuel: {self.fuel_target.fuel}')
+            print(f'  file fuel: {self.fuel_file.fuel}')
+            return self.target_goal
+
+        # Found no goal in any file, but `count_unsafe2` still reported some
+        # unsafe code.  Tell the agent to check the entire codebase for
+        # leftover unsafety.
+        print('PickTarget: no valid targets')
```
thedataking, line 578: do you need to give back the fuel you took at the top of function here?
do we expect target goal to be none here or is it okay if it is stale?

crisp/__main__.py
```diff
@@ -491,15 +495,80 @@ def safety_loop_common(args, cfg, mvir, w, n_code, n_c_code):
...
+        self.fuel_file = FuelCounter('target file',
+            default_give=limits.safety_tries_per_file)
+        self.fuel_target = FuelCounter('target goal',
+            default_give=limits.safety_tries_per_target)
+        self.file_target_goals = []
+        self.current_file = None
+        self.target_goal = None
+
+    def current_target_goal(self, w, n_code):
+        # Keep going with the current goal if possible.
```
thedataking, line 523: tiny nit: consider commenting on the try-use-then-give pattern. otherwise the user has to trace the whole function to convince themselves there is no double spending.

tools/find_unsafe2/src/lib.rs
```diff
@@ -179,6 +179,8 @@ pub struct Outputs {
 pub struct FunctionOutputs {
     /// Sum of all unsafe counts for this function.
     pub total_unsafe: usize,
+    /// Name of the file that contains this function.
+    pub filename: String,
```
thedataking, line 183: I guess this breaks existing runs... not sure we care.

thedataking (approved):

spernsteiner:
Updated the submodule to include https://github.com/immunant/c2rust/pull/1902

crisp/workflow.py
spernsteiner: Added a comment in 6d1d8af29eefa60dc16a338ced9c674f73e608ad

crisp/workflow.py
spernsteiner: Added `.replace('#![feature(raw_rep_op)]', '')` in 6d1d8af29eefa60dc16a338ced9c674f73e608ad

crisp/__main__.py
spernsteiner: Modified this in 6d1d8af29eefa60dc16a338ced9c674f73e608ad to set `self.target_goal` and give back fuel for consistency with other cases

crisp/__main__.py
spernsteiner: Added a comment in 6d1d8af29eefa60dc16a338ced9c674f73e608ad noting that fuel is being taken optimistically here and that fuel counters will be reset below if they turn out to be empty

tools/find_unsafe2/src/lib.rs
spernsteiner: Ideally analysis caching would detect when the tool binary has changed and invalidate old cached results automatically.  But for now, during development I've been adding `CACHE_BUSTER=123` to the env vars of the command, which has no effect on the output but changes part of the inputs that are used as the cache key.

crisp/workflow.py
spernsteiner: Added a link to #146 in 378b8dfb9

PR #155, opened 2026-07-09
Make the agent's life a little bit easier

randomPoison:
This PR makes a couple of changes to smooth out some speedbumps I see the agent hitting regularly:

- **Give the agent a way to diff the current workspace** against the original code for the current turn, and also tell it that it's not in a Git repo. The agent regularly tries to `git diff` to see its pending changes, which doesn't work because it's not in a Git repo. This PR adds an additional copy of the starting code for the turn in a separate directory, and adds instructions to tell the agent how it can diff against that baseline to check its pending changes.
- **Explain more of how the unsafe JSON works** and how to interact with it.
  - Explain where to find the baseline JSON and that check-unsafe2 does not update that file.
  - Tell it how to use find-unsafe2 to generate updated unsafety counts without overwriting the baseline JSON.
  - Tell it how to avoid dumping the raw JSON into context, including avoiding common ways it accidentally dumps the full JSON e.g. running `rg` on the whole workspace.
  - Give it advice on how to use `jq` to more cleanly query the unsafety JSON.

These changes should somewhat help reduce unnecessary token usage and help the agent work a bit faster. I doubt this will have any impact on performance, at least wrt the quality of the translated code, but it should reduce token usage and help the agent move a bit more quickly. I don't see a massive difference on cJSON, but I think this will have a bigger impact on larger projects, where dumping the full unsafety JSON into context can be a pretty massive amount of context flooding.

Also, the instructions on how to avoid overwriting the unsafety JSON should help reduce the number of rejected edits. When I tested these changes in combination with the fix in https://github.com/GaloisInc/Tractor-Crisp/pull/151, I was able to do two full translations of cJSON with 0 rejected edits.

### DIFF
--- a/crisp/agent.py
+++ b/crisp/agent.py
@@ -5,6 +5,7 @@ Rewrite operations using AI agent tools, such as codex-cli
 import os
 from pathlib import Path
 import re
+import shlex
 
 from pathspec.pathspec import PathSpec
 
@@ -15,6 +16,7 @@ from .mvir import MVIR, TreeNode, FileNode, CodexAgentOpNode
 from .sandbox import run_sandbox
 
 AGENT_DEFAULT_MODEL = "gpt-5.5-2026-04-23"
+BASELINE_DIR = '/tmp/crisp-baseline'
 
 _SNAPSHOT_SUFFIX = re.compile(r"^(?P<alias>.+)-\d{4}-\d{2}-\d{2}$")
 
@@ -96,6 +98,29 @@ def _inject_codex_auth(sb):
     sb.checkout_file_untracked('.codex/auth.json', auth_bytes)
 
 
+def _normalize_baseline_path(path: str) -> str:
+    path = os.path.normpath(path)
+    if os.path.isabs(path) or path.startswith(os.pardir + os.sep) or path == os.pardir:
+        raise ValueError(f'baseline path must be relative to the sandbox work dir: {path!r}')
+    return path
+
+
+def _with_baseline_setup(sb, cmd: list[str], baseline_paths: list[str]) -> list[str]:
+    if len(baseline_paths) == 0:
+        return cmd
+
+    script_lines = [f'rm -rf {shlex.quote(BASELINE_DIR)}']
+    for path in baseline_paths:
+        path = _normalize_baseline_path(path)
+        src = sb.join(path)
+        dest = os.path.join(BASELINE_DIR, path)
+        script_lines.append(f'mkdir -p {shlex.quote(os.path.dirname(dest))}')
+        script_lines.append(f'cp -a {shlex.quote(src)} {shlex.quote(dest)}')
+    script_lines.append(f'exec {shlex.join(cmd)}')
+
+    return ['sh', '-c', ' && '.join(script_lines)]
+
+
 def run_rewrite(
     cfg: Config,
     mvir: MVIR,
@@ -108,6 +133,7 @@ def run_rewrite(
     codex_login: bool = False,
     env: dict | None = None,
     find_unsafe2_json_dir: str | None = None,
+    baseline_paths: list[str] | None = None,
 ) -> tuple[TreeNode, TreeNode]:
     # Print the warning in red so it stands out
     WARNING_TEMPLATE = "\033[31mwarning: {} is being copied into " \
@@ -124,6 +150,8 @@ def run_rewrite(
 
     if env is None:
         env = {}
+    if baseline_paths is None:
+        baseline_paths = []
 
     with run_sandbox(cfg, mvir) as sb:
         sb.checkout(input_code)
@@ -144,6 +172,7 @@ def run_rewrite(
             '--skip-git-repo-check',
             prompt,
         ], codex_login=codex_login)
+        codex_cmd = _with_baseline_setup(sb, codex_cmd, baseline_paths)
         print(codex_cmd)
         all_cmds = [mkdir_codex, codex_cmd] + clean_cmds
 
--- a/crisp/workflow.py
+++ b/crisp/workflow.py
@@ -162,12 +162,14 @@ cargo check-unsafe2 --manifest-path {cargo_dir_path}/Cargo.toml
 ```
 This will report an error for any unsafe code that was improperly added during your edits. It also reports errors on any newly added "unsafe-adjacent" code, including int-to-pointer casts and arguments or fields of raw pointer type.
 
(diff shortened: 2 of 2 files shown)


randomPoison:
The way Codex wrote the logic for creating the baseline copy of the current code feels a little weird to me, but I've tested out the current code and it seems to work fine. I can tweak that further if anyone thinks the implementation is wrong, but I'm inclined to leave it as-is so that I don't have to do another set of test runs.

crisp/agent.py
```diff
@@ -96,6 +98,29 @@ def _inject_codex_auth(sb):
     sb.checkout_file_untracked('.codex/auth.json', auth_bytes)
 
 
+def _normalize_baseline_path(path: str) -> str:
+    path = os.path.normpath(path)
+    if os.path.isabs(path) or path.startswith(os.pardir + os.sep) or path == os.pardir:
```
ahomescu, line 103: There is `relative_to` in `pathlib`, would that do what you need here? If so, it's probably better to use since it's more tested.

crisp/agent.py
```diff
@@ -96,6 +98,29 @@ def _inject_codex_auth(sb):
...
+        raise ValueError(f'baseline path must be relative to the sandbox work dir: {path!r}')
+    return path
+
+
+def _with_baseline_setup(sb, cmd: list[str], baseline_paths: list[str]) -> list[str]:
+    if len(baseline_paths) == 0:
+        return cmd
+
+    script_lines = [f'rm -rf {shlex.quote(BASELINE_DIR)}']
```
ahomescu, line 112: Could you use `unlink` from `pathlib` here instead?

crisp/workflow.py
```diff
@@ -149,11 +149,49 @@
...
 ```
 This will report an error for any unsafe code that was improperly added during your edits. It also reports errors on any newly added "unsafe-adjacent" code, including int-to-pointer casts and arguments or fields of raw pointer type.
+
+The baseline unsafe analysis JSON for the start of this turn is available under `unsafe_json/*.json`, and the absolute directory is in `$FIND_UNSAFE2_JSON_DIR`. `cargo check-unsafe2` uses this baseline JSON to compare your edited code against the start-of-turn code. Do not delete, overwrite, or regenerate files under `unsafe_json/`, because that can make `cargo check-unsafe2` compare your edits against themselves and miss newly introduced unsafe code.
+
+When searching source text with `rg` or similar tools, search inside `{cargo_dir_path}/` rather than from the workspace root, for example `rg "PATTERN" {cargo_dir_path}` or `rg "PATTERN" {cargo_dir_path}/src`. Broad workspace-level searches can accidentally match the large `unsafe_json/*.json` files and flood your transcript. If you need information from `unsafe_json/*.json`, use the targeted `jq` queries below instead of text search.
```
ahomescu, line 167: Or we could create a `.gitignore` file and put `unsafe_json` there, afaict ripgrep ignores files listed there. It also supports a more generic `.ignore` file, and its own `.rgignore`.

ahomescu (changes requested):

crisp/agent.py
```diff
@@ -15,6 +16,7 @@
 from .sandbox import run_sandbox
 
 AGENT_DEFAULT_MODEL = "gpt-5.5-2026-04-23"
+BASELINE_DIR = '/tmp/crisp-baseline'
```
ahomescu, line 19: Do we need to put it in `/tmp`? Couldn't it be in a separate subdirectory in the workspace, like `original_rust`?

crisp/workflow.py
```diff
@@ -149,11 +149,49 @@
...
+Replace `path/to/file.rs` with the relative path of the source file you edited. Avoid directory-level diffs after running `cargo build`, because generated files under `target/` can flood the output.
+Do not edit files under `/tmp/crisp-baseline`, as this will break your ability to diff cleanly.
+
 Your changes must not introduce new unsafe code within implementation functions. You can check your work using this command:
 ```sh
 cargo check-unsafe2 --manifest-path {cargo_dir_path}/Cargo.toml
 ```
 This will report an error for any unsafe code that was improperly added during your edits. It also reports errors on any newly added "unsafe-adjacent" code, including int-to-pointer casts and arguments or fields of raw pointer type.
+
+The baseline unsafe analysis JSON for the start of this turn is available under `unsafe_json/*.json`, and the absolute directory is in `$FIND_UNSAFE2_JSON_DIR`. `cargo check-unsafe2` uses this baseline JSON to compare your edited code against the start-of-turn code. Do not delete, overwrite, or regenerate files under `unsafe_json/`, because that can make `cargo check-unsafe2` compare your edits against themselves and miss newly introduced unsafe code.
```
ahomescu, line 165: You could split the unsafe JSON changes into a separate PR.

crisp/workflow.py
ahomescu: Alternatively, if you want the agent to figure it all out instead instead of feeding it the ignore file, you could tell it specifically "don't look under `unsafe_json`" instead. My objection is to the `search inside [...]` part.

randomPoison: I don't think just having a `.gitignore` is enough. I'm currently running the `git init` experiment, and I included a `.gitignore` in there that includes ignoring `unsafe_json/`, but the agent is still deliberately trying to `rg` instead the `unsafe_json/` dir. I think we need to specifically tell the agent that the unsafe JSON shouldn't be grepped, and it's probably useful to warn it about ways it can accidentally dump the JSON

spernsteiner: Could we instead put the unsafe JSON under `/tmp` or somewhere where the agent isn't likely to stumble across it?  `check-unsafe2` expects `$FIND_UNSAFE2_JSON_DIR` to be an absolute path, so there's no constraint on where we put it.

randomPoison: Sure, I think that'd at least help the agent not accidentally grep through the JSON when it searches within the workspace. But the agent still wants to query the JSON sometimes, so it's going to need to know where the JSON is, and it's going to try to grep through that file unless we tell it otherwise. I ultimately don't care where we put the JSON, but no matter where we put it we're going to have to tell the agent where it is and how to interact with it.

randomPoison:
https://github.com/GaloisInc/Tractor-Crisp/pull/159 has an alternate fix for the `git diff` issue, instead opting to `git init` a repo within the sandbox that the agent can then use.

randomPoison:
#159 landed with an alternate solution for giving the agent a git repo, so this PR is no longer entirely relevant. I'm going to close it since I'm not actively working on TRACTOR, but someone else with time should try to land some kind of fix for the issue of dumping the unsafe JSON into context, since that's likely a big source of context pollution in larger projects.

PR #156, opened 2026-07-13
crisp: add model selection configuration

thedataking:
- existing `models` configuration key becomes `model_options`
- also changes agent default to gpt-5.6-terra  which is better and cheaper than the current gpt-5.5 default.
- updated README and `crisp.toml.example`

We need per-task/stage model selection for #131 and it would also be good for #152

### DIFF
--- a/crisp/config.py
+++ b/crisp/config.py
@@ -39,6 +39,13 @@ class ConfigBase:
         return cls.from_dict(d, path, **kwargs)
 
 @dataclass(frozen = True)
+class ModelsConfig(ConfigBase):
+    agent: str  = "gpt-5.6-terra"
+    postprocess: str = "gpt-5.6-luna"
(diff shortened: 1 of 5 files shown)


crisp/config.py
```diff
@@ -38,6 +38,13 @@ def from_toml_file(cls, f, **kwargs):
         d = toml.load(f)
         return cls.from_dict(d, path, **kwargs)
 
+@dataclass(frozen = True)
+class ModelsConfig(ConfigBase):
+    agent: str  = "gpt-5.6-terra"
+    postprocess: str = "gpt-5.6-luna"
```
spernsteiner, line 44: I was going to suggest pinning to a specific model snapshot, like how we were previously on `gpt-5.5-2026-04-23`, but apparently there's no such thing for GPT-5.6

spernsteiner (approved):

PR #159, opened 2026-07-14
Give the agent a little Git repo in the sandbox

randomPoison:
I've observed the agent trying to run `git diff` and a couple of other Git operations in the sandbox, which currently does not work because we don't give the agent a Git repo. https://github.com/GaloisInc/Tractor-Crisp/pull/155 attempts to fix this by telling the agent it's not in a Git repo and giving it alternate instructions for diffing the state of the code. @thedataking suggested that we could instead init a Git repo within the sandbox so that the agent could use the Git commands it's already comfortable with. That wouldn't expose the full MVIR history to the agent, but would allow things like `git diff` to work. I tested out this out and looks like the agent mainly tries to do `git diff` and `git restore`, I haven't seen it trying to search through the commit history or do other operations that wouldn't work.

One other minor fix in here related to #155 is that I've included the unsafe JSON dir in the `.gitignore` to see if that helps the agent to not accidentally dump. This doesn't seem to matter in practice since the agent explicitly points `rg` at the `unsafe_json` dir, overriding the `.gitignore` 🤷

### DIFF
--- a/crisp/agent.py
+++ b/crisp/agent.py
@@ -132,6 +132,23 @@ def run_rewrite(
         if planning_files is not None:
             sb.checkout(planning_files)
 
+        # Each agent turn gets a fresh sandbox, so make the checked-out files
+        # the baseline of a small local repository.  This lets the agent use
+        # ordinary Git commands (especially `git diff`) to inspect its edits.
+        # `git commit -a` alone does not include the initially untracked files,
+        # hence the explicit add before creating the baseline commit.
+        init_git_repo = [
+            'sh', '-c',
+            'git init -q && git add --all && '
+            'git -c user.name=CRISP -c user.email=crisp@localhost '
+            'commit --quiet -m "CRISP sandbox baseline"',
+        ]
+        exit_code, logs = sb.run(init_git_repo, cwd=cwd, stream=True)
+        if exit_code != 0:
+            raise CrispError(
+                f'failed to initialize Git repository in agent sandbox: '
+                f'exit code {exit_code}')
+
         if codex_login:
             print(WARNING_TEMPLATE.format('codex\'s login session (`auth.json`)'))
             _inject_codex_auth(sb)
@@ -163,6 +180,7 @@ def run_rewrite(
                 raise CrispError(f'codex-cli failed: exit code {exit_code}')
 
         ignore_lines = [
+            '.git/',
             '__pycache__/',
             'build/',
             'build-ninja/',
--- a/crisp/agent.py
+++ b/crisp/agent.py
@@ -5,6 +5,7 @@ Rewrite operations using AI agent tools, such as codex-cli
 import os
 from pathlib import Path
 import re
+import shlex
 
 from pathspec.pathspec import PathSpec
 
@@ -137,9 +138,23 @@ def run_rewrite(
         # ordinary Git commands (especially `git diff`) to inspect its edits.
         # `git commit -a` alone does not include the initially untracked files,
         # hence the explicit add before creating the baseline commit.
+        gitignore_lines = [
+            '# Cargo build output',
+            '/target/',
+            '**/*.rs.bk',
+            '*.pdb',
+        ]
+        if find_unsafe2_json_dir is not None:
+            gitignore_lines.extend([
+                '# CRISP unsafe-analysis results',
+                f'{find_unsafe2_json_dir.rstrip("/")}/',
+            ])
+        write_gitignore = (
+            f'printf %s {shlex.quote("\\n".join(gitignore_lines) + "\\n")} '
+            '> .gitignore')
         init_git_repo = [
             'sh', '-c',
-            'git init -q && git add --all && '
+            f'git init -q && {write_gitignore} && git add --all && '
             'git -c user.name=CRISP -c user.email=crisp@localhost '
             'commit --quiet -m "CRISP sandbox baseline"',
         ]
--- a/crisp/agent.py
+++ b/crisp/agent.py
@@ -5,7 +5,6 @@ Rewrite operations using AI agent tools, such as codex-cli
 import os
 from pathlib import Path
 import re
-import shlex
 
 from pathspec.pathspec import PathSpec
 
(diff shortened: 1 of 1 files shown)


crisp/agent.py
```diff
@@ -130,14 +133,45 @@ def run_rewrite(
...
+            '*.pdb',
+        ]
+        if find_unsafe2_json_dir is not None:
+            gitignore_lines.extend([
+                '# CRISP unsafe-analysis results',
+                f'{find_unsafe2_json_dir.rstrip("/")}/',
+            ])
+        write_gitignore = (
+            f'printf %s {shlex.quote("\\n".join(gitignore_lines) + "\\n")} '
+            '> .gitignore')
```
spernsteiner:
```suggestion
        gitignore_file = FileNode.new(mvir, '\n'.join(gitignore_lines))
        sb.checkout_file('.gitignore', gitignore_file)
```

crisp/agent.py
```diff
@@ -130,14 +133,45 @@ def run_rewrite(
...
+        init_git_repo = [
+            'sh', '-c',
+            f'git init -q && {write_gitignore} && git add --all && '
```
spernsteiner: `sb.run` takes a `shell=True` kwarg so you don't need to wrap things in `sh -c` manually.

crisp/agent.py
```diff
@@ -130,14 +133,45 @@ def run_rewrite(
...
             sb.checkout(planning_files)
 
+        # Each agent turn gets a fresh sandbox, so make the checked-out files
+        # the baseline of a small local repository.  This lets the agent use
+        # ordinary Git commands (especially `git diff`) to inspect its edits.
+        # `git commit -a` alone does not include the initially untracked files,
+        # hence the explicit add before creating the baseline commit.
+        gitignore_lines = [
+            '# Cargo build output',
+            '/target/',
```
spernsteiner: I think the target dir will usually end up being something like `translated_rust/target/` (relative to the agent's `cwd`) rather than `target/`.  For example, the `clar` test support in `test_eval.py` includes in the test command the line `cargo build --manifest-path translated_rust/Cargo.toml`, which will put the target dir in `translated_rust/target/`.

crisp/agent.py
```diff
@@ -130,14 +133,45 @@ def run_rewrite(
...
+            '**/*.rs.bk',
+            '*.pdb',
```
spernsteiner: What produces .rs.bk or .pdb files?  I've never seen these (aside from .pdb for windows debug info, which shouldn't be relevant here).

crisp/agent.py
```diff
@@ -130,14 +133,45 @@ def run_rewrite(
...
+        write_gitignore = (
+            f'printf %s {shlex.quote("\\n".join(gitignore_lines) + "\\n")} '
+            '> .gitignore')
+        init_git_repo = [
+            'sh', '-c',
+            f'git init -q && {write_gitignore} && git add --all && '
+            'git -c user.name=CRISP -c user.email=crisp@localhost '
+            'commit --quiet -m "CRISP sandbox baseline"',
+        ]
+        exit_code, logs = sb.run(init_git_repo, cwd=cwd, stream=True)
```
spernsteiner: Is `cwd=cwd` correct for the git setup step?  We pass it when running `codex`, though currently no call sites set `cwd` to anything other than the default (`.`).  If it was set, do we want the git repo to be limited to only the codex subdirectory (in which case some paths above may need adjusting)?  Or do we want it to cover the whole sandbox (which might confuse codex, having the git root outside its work directory)?

crisp/agent.py
```diff
@@ -130,14 +133,45 @@ def run_rewrite(
...
+        if exit_code != 0:
+            raise CrispError(
+                f'failed to initialize Git repository in agent sandbox: '
+                f'exit code {exit_code}')
```
spernsteiner: Ideally we should wait to raise the exception until after we've built the `CodexAgentOpNode` (contrary to the way it's done elsewhere in this function), so we have logs in `crisp-storage` recording the bad agent invocation.

spernsteiner (commented):
Might be worth clarifying that this is intended to commit the entire sandbox directory into git, not just the `translated_rust/` subdirectory

crisp/agent.py
randomPoison: Oh, there's probably nothing generating those files. I just asked Codex to give me a default `.gitignore` for a Rust project, I think those files get generated on Windows systems (at least the `.pdb`). I'll simplify that down to just ignoring `target/`.

crisp/agent.py
randomPoison: Setting `cwd=cwd` seems to have worked fine, I saw no issues with the agent interacting with the repo nor any indication that including the whole sandbox caused any confusion. I think including the whole sandbox in the repo is what we want, specifically including the unsafety JSON is useful because sometimes the agent overwrites the JSON and this gives it a way to revert that change. So I think `cwd="."` is what we want here.

crisp/agent.py
randomPoison: Switched this to ignore `target/` which I think should ensure that `target` gets ignored even in a subdirectory.

randomPoison:
> Might be worth clarifying that this is intended to commit the entire sandbox directory into git, not just the `translated_rust/` subdirectory

Updated the comment in there to explicitly call that out. The PR is updated to address the feedback, I'm going to do another test run to verify that everything works with these changes.

randomPoison:
Seems like the updated code is working, though this run finally turned up evidence that the agent sometimes wants to `git log` and `git blame` (which I haven't seen previously):

<details><summary>Codex's analysis of the Git operations performed by the agent</summary>

> Mostly yes—the synthetic Git repository is functioning correctly, but this run exposed two limitations worth noting.
> 
>   - git status, diff, diff --check, checkout, blame, and patch reversal all operated against the sandbox repository normally.
>   - The agent intentionally used git checkout -- rust/src/cJSON.rs several times to abandon unsuccessful current-iteration experiments. Because HEAD is the current CRISP
>     node’s baseline, these restores preserved previously accepted CRISP work.
> 
>   - It successfully reversed one large incomplete edit by saving its diff and applying it with git apply -R. with_tests_2/run_20260715_134717.log:25961
>   - It never committed, branched, reset the repository, or modified remotes.
>   - .codex/ and build-rust-tests/ appeared as untracked files in status output, but they were never included in diffs or added to the repository.
> 
>   Two things were imperfect:
> 
>   1. History queries were syntactically successful but uninformative. The agent ran git log three times and git blame once. Each invocation could see only that sandbox’s
>      single CRISP sandbox baseline commit. with_tests_2/run_20260715_134717.log:12022 with_tests_2/run_20260715_134717.log:17773
> 
>      Thus Git history did not expose upstream cJSON history or earlier CRISP edits. The agent did not make any consequential decision based on the misleading history; it
>      mostly used the source itself.
> 
>   2. Several handcrafted git apply patches failed. There were five “corrupt patch” errors and two context-mismatch failures. with_tests_2/run_20260715_134717.log:4348 These
>      failures came from malformed patch text, not from a broken repository. The agent noticed them and reapplied the edits through Python, Ruby, or Perl.
> 
>      Some command groups lacked set -e, so the overall shell invocation was reported as successful even though git apply printed an error. That is the main thing that looks
>      genuinely awkward, although it did not corrupt or lose any final work.
> 
>   Overall, the repo is doing its primary job well: status, diffs, cleanup, and restoration work. The only feature it cannot meaningfully provide is real historical context,
>   and the agent did try to use that context in this run.

</details>

~@scuellar~ @spernsteiner  Let me know if you want any more changes on this before I merge.

crisp/agent.py
```diff
@@ -130,37 +130,66 @@ def run_rewrite(
...
+
+            exit_code = 0
+            logs = b''
+            for cmd in all_cmds:
+                exit_code, logs2 = sb.run(cmd, cwd=cwd, stream=True, env=env)
+                logs += logs2
+
+                # TODO: ensure API key doesn't get included in the AgentOpNode
+                if exit_code != 0:
+                    raise CrispError(f'codex-cli failed: exit code {exit_code}')
```
spernsteiner: Since the other exception was moved out to after building the op, this one should be too:


```suggestion
            logs = b''
            for cmd in all_cmds:
                exit_code, logs2 = sb.run(cmd, cwd=cwd, stream=True, env=env)
                logs += logs2

                # TODO: ensure API key doesn't get included in the AgentOpNode
                if exit_code != 0:
                    break
```

crisp/agent.py
```diff
@@ -130,37 +130,66 @@ def run_rewrite(
...
+            mvir, '\n'.join(gitignore_lines) + '\n')
+        sb.checkout_file('.gitignore', gitignore_file)
+        init_git_repo = (
+            'git init -q && git add --all && '
+            'git -c user.name=CRISP -c user.email=crisp@localhost '
+            'commit --quiet -m "CRISP sandbox baseline"')
+        exit_code, logs = sb.run(init_git_repo, cwd='.', shell=True, stream=True)
+        git_init_failed = exit_code != 0
+
+        if not git_init_failed:
```
spernsteiner:
```suggestion

        if exit_code == 0:
```

crisp/agent.py
```diff
@@ -218,4 +247,9 @@ def run_rewrite(
     # Record operations and timestamps in the `op_history` reflog.
     mvir.set_tag('op_history', n_op.node_id(), n_op.kind)
 
+    if git_init_failed:
```
spernsteiner:
```suggestion
    if exit_code != 0:
```

crisp/agent.py
```diff
@@ -218,4 +247,9 @@ def run_rewrite(
...
+        raise CrispError(
+            f'failed to initialize Git repository in agent sandbox: '
+            f'exit code {exit_code}')
```
spernsteiner: `CrispError` can take a relevant `Node` to record which operation specifically failed:
```suggestion
        raise CrispError(
            f'failed to initialize Git repository in agent sandbox: '
            f'exit code {exit_code}', n_op)
```

spernsteiner (approved):
Looks good.

Comments are some specific suggestions about making the exception handling consistent (not sure if I got the details exactly right, so please check).  The usual pattern in other code is to do something like this:
```python
exit_code, logs = sb.run(first_command)

if exit_code == 0:
    exit_code, logs2 = sb.run(second_command)
    logs += logs2

if exit_code == 0:
    exit_code, logs2 = sb.run(third_command)
    logs += logs2
```
Then at the end build some kind of `Node` containing the `exit_code` and `logs`, and throw an exception if `exit_code != 0`.

PR #167, opened 2026-07-25
Replace per-edit unsafety ratchet with reviewed net-decrease policy

thedataking:
`check-unsafe2` enforced a ratchet per function and per metric: no unsafety metric could increase anywhere, even when the crate-wide result improved. That rule could not distinguish adding unsafe code from relocating it into a smaller, explicit boundary, so legitimate decompositions were rejected in flight and the rewriter often spent its budget reverting them or landing no-ops.

The new policy judges the bounded net effect, reviews what counting cannot decide, and gives stalled runs an honest way to stop:

- Functions free of unsafe remain protected. Existing unsafe functions, and new named implementation functions created as part of a relocation, may absorb unsafe operations only when the crate-wide total does not increase. Raw-pointer fields, int-to-pointer casts, new foreign-function uses, and calls into FFI entry points remain hard errors. Pointer-bearing signature increases are permitted on `unsafe fn`, while a new or newly-safe function must have a pointer-free signature.
- Every checker-tolerated per-function increase is sent to a dedicated review with an anti-laundering rubric. FFI entry-point changes retain their separate review, and the checker now warns immediately when work grows inside a count-exempt wrapper. Hard checker failures run before expensive tests or reviews.
- The adversarial corpus covers net increases, unsafe relocation into new helpers, reference-oracle laundering, raw-pointer signature flips, unsafe helpers used for decomposition, and implementation work stuffed into FFI wrappers.
- The loop can accept a net-neutral preparatory step or a final `BLOCKED:` result instead of pressuring the rewriter into cosmetic or rule-breaking edits. Dead-end notes must cite the exact checker diagnostic, review finding, or failing test, and oversized plans can be compacted without changing their remaining work.
- Near the end of a stalled run, a read-only adjudicator can waive residual operations only when they are sound, minimal, and inherent to the FFI boundary. Planning, adjudication, and review have separate model and effort settings, and a run whose remaining operations are all covered ends explicitly as "completed with residual unsafe."


thedataking:
Superseded by #182, stacked on the cleaned-up #171. The replacement carries the current safety-loop work and uses the checker policy on main; the relaxed checker policy and residual-unsafety waiver experiments from this PR are not included.

PR #169, opened 2026-07-29
find_unsafe2: harden unsafe regression checks

thedataking:
The safety loop's rewriter optimizes against check-unsafe2's metrics, and run logs show it probing their boundaries: laundering helpers and entry-point minting attempts in a zlib run; relocation of pointer work into exempt wrappers and ptr→int distance arithmetic in a libexpat run. Each change here either closes an evasion channel (making a metric follow the unsafety wherever a rewrite can move it) or fixes a counting artifact that rejected legitimate rewrites:

 - Attribute closure metrics to the enclosing function: closure keys are positional (`{closure#N}`) and rekey on unrelated edits, so as standalone baseline entries they reject unchanged code. 28% of checker diagnostics in the zlib run were rekeyed closures, none a real regression. Folding also means a closure can't mint a fresh all-zero budget.
 - Look through local structs in sig_contains_raw_ptr: wrapping a pointer in a local struct would zero the signature count with no real progress.
 - Charge pointer-materializing std calls (with_exposed_provenance & friends) as int-to-ptr casts: the same round-trip the as-cast metric already blocks.
 - Charge inline asm as unsafety: arbitrary memory access, previously uncounted.
 - Charge impl-code uses of local FFI entry points: entry points are metric-exempt, so routing work through them hides unsafe code; enforces the existing FFI rule deterministically instead of relying on the LLM review.
 - Track exported symbols and reject export-set changes: deleting or demoting an exported fn/static, or editing an `#[export_name]` value, was invisible to the checker; the symbol set is the ABI.
 - Reject new unsafe impl: a zero-code soundness assertion no body metric sees.
 - Hold crates missing from the baseline to an empty inventory, so unsafe can't hide in a newly added crate.
 - Adversarial test suite pinning these accepted/rejected rewrites as snapshots


tools/find_unsafe2/src/bin/check_unsafe2.rs
```diff
@@ -84,25 +138,33 @@ fn check_function_outputs(name: &str, old: &FunctionOutputs, new: &FunctionOutpu
...
-    ok &= check_bad_flag(old.is_ffi_entry_point, is_ffi_entry_point,
-        || format!("{name}: FFI export flag"));
-
     ok
 }
 
+fn fmt_symbol(sym: &Option<String>) -> String {
+    match sym {
+        Some(s) => format!("{s:?}"),
+        None => String::from("none"),
```
ahomescu: Does this collide with a symbol named `none`? Maybe use `<none>` here instead.

tools/find_unsafe2/src/lib.rs
```diff
@@ -194,6 +224,9 @@ pub struct FunctionOutputs {
     pub derefs_raw_ptr: usize,
     /// Unsafety: function calls `unsafe fn`s.
     pub calls_unsafe: usize,
+    /// Unsafety: function contains inline assembly, which can access arbitrary memory.
+    #[serde(default)]
```
ahomescu: Why is this needed here?

tools/find_unsafe2/src/lib.rs
```diff
@@ -212,6 +245,13 @@ pub struct FunctionOutputs {
 
     /// Progress: function mentions imported `extern` `fn`s.
     pub uses_foreign_fn: IndexMap<String, usize>,
+    /// Progress: function mentions local FFI entry points.
+    ///
+    /// The FFI rules forbid implementation code from calling or referencing exported entry
+    /// points, and entry points are exempt from the unsafety checks, so routing work through
+    /// them would hide unsafe code.
+    #[serde(default)]
```
ahomescu: Same here: the other fields don't have this attribute.

tools/find_unsafe2/src/lib.rs
```diff
@@ -405,6 +555,25 @@ pub fn process(tcx: TyCtxt) -> Outputs {
         }
     };
 
+    let mut is_fn_ffi_entry_point = {
+        let mut storage = HashMap::new();
+        move |fd: FnDef| -> bool {
+            if let Some(&x) = storage.get(&fd) {
+                return x;
+            }
+            let internal_def_id = rustc_internal::internal::<DefId>(tcx, fd.0);
```
ahomescu: What's this for, why not pass `fd.0` directly to `is_foreign_item`? This needs a comment.

tools/find_unsafe2/src/lib.rs
```diff
@@ -419,46 +588,70 @@ pub fn process(tcx: TyCtxt) -> Outputs {
...
+    {
+        use rustc_hir::def::DefKind;
+        for item_id in tcx.hir_free_items() {
+            let did = item_id.owner_id.to_def_id();
+            if !matches!(tcx.def_kind(did), DefKind::Impl { of_trait: true }) {
+                continue;
+            }
+            let header = tcx.impl_trait_header(did);
+            if matches!(header.safety, rustc_hir::Safety::Unsafe) {
+                *out.unsafe_impls.entry(tcx.def_path_str(did)).or_insert(0) += 1;
```
ahomescu: Nit: `.or_insert(0)` could be just `.or_default()`.

ahomescu (changes requested):

crisp/analysis.py
```diff
@@ -507,6 +507,8 @@ def _check_unsafe2_impl(cfg: Config, mvir: MVIR, sb: Sandbox,
     )
     return n_op
 
+# TODO: reject baseline crates that produced no JSON, so dropping a crate from
+# the build can't silently retire its unsafety
```
spernsteiner: What does this mean?  Every crate in the workspace should produce json (though that happens in `find_unsafe2`, not here in `check_unsafe2`), and dropping an unsafe crate from the workspace seems like a fine way to reduce unsafe if you can manage it.  And I'm not sure how the two things are supposed to relate.

crisp/analysis.py
```diff
@@ -515,6 +517,7 @@ def check_unsafe2(cfg: Config, mvir: MVIR,
     with run_sandbox(cfg, mvir) as sb:
         cmd = [
             'env',
+            f'FIND_UNSAFE2_SRC_DIR={sb.join(cargo_dir)}',
```
spernsteiner, line 518: Why is this necessary?  The `find_unsafe2` cargo subcommands default `FIND_UNSAFE2_SRC_DIR` to `.`, which should be the same as `cargo_dir`.

tools/find_unsafe2/src/bin/check_unsafe2.rs
```diff
@@ -21,9 +21,20 @@ use find_unsafe2::{self, Outputs, FunctionOutputs, TypeOutputs};
...
-    let Outputs { total_unsafe: _, ref fns, ref types } = *new;
+    let Outputs { total_unsafe: _, ref fns, ref types, ref unsafe_impls } = *new;
     let mut ok = true;
 
+    // A new `unsafe impl` asserts a soundness property no metric can verify.  Legitimate uses
+    // exist (e.g. `Send` on an owned-state registry), so this is a candidate for review-gated
+    // relaxation once a reviewer exists; until then it is an error.
+    for (k, &new_count) in unsafe_impls {
+        let old_count = old.unsafe_impls.get(k).copied().unwrap_or(0);
+        if new_count > old_count {
```
spernsteiner: The comment makes it sound like something special is happening here.  How is this check different from the usual "count must not increase" checks we do everywhere else?

tools/find_unsafe2/src/bin/check_unsafe2.rs
```diff
@@ -34,16 +45,57 @@ fn check_outputs(old: &Outputs, new: &Outputs) -> bool {
...
+        }
+    }
+
     for (fn_name, new_fn) in fns {
-        let old_fn = old.fns.get(fn_name).unwrap_or(&empty_fn);
-        ok &= check_function_outputs(fn_name, old_fn, new_fn);
+        match old.fns.get(fn_name) {
+            Some(old_fn) => {
+                if old_fn.ffi_symbol != new_fn.ffi_symbol {
+                    println!("{fn_name}: exported symbol changed: {} -> {}",
```
spernsteiner: What is this doing?  I like the idea of asserting that exported symbols weren't removed, since there may be exports not exercised by the test suite that should still be preserved, but the changes here seem very convoluted and I'm not sure why.

tools/find_unsafe2/src/bin/check_unsafe2.rs
```diff
@@ -156,18 +218,36 @@ fn main() {
...
+        if crate_name == "build_script_build" {
+            // Build scripts are skipped, matching `find_unsafe2`.
+            return ControlFlow::<(), ()>::Continue(());
+        }
+
         let json_path = json_dir.join(format!("{crate_name}.json"));
         if !fs::exists(&json_path).unwrap() {
+            // Fail closed on project crates missing from the baseline, so unsafe code can't be
+            // hidden in a newly added crate.  Crates outside `SRC_DIR` (dependencies) are skipped.
+            if find_unsafe2::any_local_item_under(src_dir) {
```
spernsteiner: Why is this an error?  Seems like we should be able to give it an empty/default `Outputs`, so it will use an empty `FunctionOutputs` for every function, and thus require every function to be free of unsafe.

tools/find_unsafe2/src/bin/check_unsafe2.rs
```diff
@@ -68,10 +120,12 @@ fn check_function_outputs(name: &str, old: &FunctionOutputs, new: &FunctionOutpu
...
         total_unsafe: _,
         filename: _,
-        is_unsafe_fn, is_mut_static, derefs_raw_ptr, calls_unsafe,
+        is_unsafe_fn, is_mut_static, derefs_raw_ptr, calls_unsafe, inline_asm,
         ref uses_static_mut, ref uses_union_field, ref uses_foreign_fn,
+        ref uses_ffi_entry_point,
         casts_int_to_ptr, sig_contains_raw_ptr,
-        is_ffi_entry_point,
+        // Checked at the crate level, where the export sets are in view.
+        is_ffi_entry_point: _, ffi_symbol: _,
```
spernsteiner: I think this comment applies only to `ffi_symbol`.  `is_ffi_entry_point` is checked just above and disables safety checks for this function.

tools/find_unsafe2/src/lib.rs
```diff
@@ -171,7 +196,12 @@ pub struct Outputs {
     pub fns: IndexMap<String, FunctionOutputs>,
     pub types: IndexMap<String, TypeOutputs>,
 
-    // TODO: Unsafety: crate implements `unsafe trait`s.
+    /// `unsafe impl` blocks (e.g. `unsafe impl Send`), keyed by impl path.  These are zero-code
+    /// soundness assertions with legitimate uses (c2rust never emits them, so any occurrence was
+    /// added by refactoring); they are surfaced for review rather than counted or rejected.
+    #[serde(default)]
+    pub unsafe_impls: IndexMap<String, usize>,
```
spernsteiner: Misleading comment.  New unsafe impls are rejected by check_unsafe2.

tools/find_unsafe2/src/lib.rs
```diff
@@ -227,6 +267,9 @@ pub struct FunctionOutputs {
     /// Whether this function is an FFI entry point.  Specifically, this is set when the function
     /// has the `#[no_mangle]` or `#[export_name = ...]` attribute.
     pub is_ffi_entry_point: bool,
+    /// The exported symbol: the `#[export_name]` value, or the item's own name under
+    /// `#[no_mangle]`.  Set for exported `fn`s and `static`s; the symbol set is the ABI.
+    pub ffi_symbol: Option<String>,
```
spernsteiner:
```suggestion
    /// The exported symbol: the `#[export_name]` value, or the item's own name under
    /// `#[no_mangle]`.  Set for exported `fn`s and `static`s.
    pub ffi_symbol: Option<String>,
```

Also, this makes `is_ffi_entry_point` redundant, since it's set exactly when `ffi_symbol` is `Some`.

tools/find_unsafe2/src/lib.rs
```diff
@@ -246,21 +289,96 @@ impl FunctionOutputs {
...
+            + inline_asm
             + uses_static_mut.values().copied().sum::<usize>()
             + uses_union_field.values().copied().sum::<usize>();
     }
+
+    /// Fold a closure's metrics into `self`, its enclosing item.  Only body-derived metrics
+    /// merge; declaration properties (qualifiers, signature, export status) are the enclosing
+    /// item's own, so a closure's raw-pointer parameters never charge the enclosing signature.
+    fn merge_from(&mut self, other: &FunctionOutputs) {
+        fn merge_map(dest: &mut IndexMap<String, usize>, src: &IndexMap<String, usize>) {
```
spernsteiner: Should use the `let FunctionOutputs { /*...*/ } = *self;` pattern with all fields listed explicitly, so we don't forget to edit this when adding new fields, and so that it's obvious which things are being excluded here.

tools/find_unsafe2/src/lib.rs
```diff
@@ -246,21 +289,96 @@ impl FunctionOutputs {
...
             + derefs_raw_ptr
             + calls_unsafe
+            + inline_asm
             + uses_static_mut.values().copied().sum::<usize>()
             + uses_union_field.values().copied().sum::<usize>();
     }
+
+    /// Fold a closure's metrics into `self`, its enclosing item.  Only body-derived metrics
+    /// merge; declaration properties (qualifiers, signature, export status) are the enclosing
+    /// item's own, so a closure's raw-pointer parameters never charge the enclosing signature.
```
spernsteiner: Why is it correct to ignore raw pointers in the closure signature?  I would think we'd want those included in the quality metrics, to encourage the agent to get rid of them.

tools/find_unsafe2/src/lib.rs
```diff
@@ -246,21 +289,96 @@ impl FunctionOutputs {
...
+        }
+    }
+    &name[..end]
+}
+
+/// Merge closure (and other unnamed-item) entries into their enclosing named functions, then
+/// recompute all totals.  Closure keys are positional (`{closure#N}`) and rekey on unrelated
+/// edits, so they must not act as standalone baseline entries; folding also lets closures
+/// created inside FFI entry points inherit the entry-point exemption.
+pub fn fold_closure_entries(out: &mut Outputs) {
```
spernsteiner: Seems cleaner to fold the visitor results into `out.fns[attribution_key(name)]` on the fly during the main loop, instead of building up a bunch of separate entries and folding them together (in what looks to be an overcomplicated manner) at the end.

tools/find_unsafe2/src/lib.rs
```diff
@@ -246,21 +289,96 @@ impl FunctionOutputs {
...
+    let fns = std::mem::take(&mut out.fns);
+    let mut folded: IndexMap<String, FunctionOutputs> = IndexMap::new();
+    // Insert named items first, so folding never depends on the order rustc emits items.
```
spernsteiner: The actual issue, not mentioned here, is that `merge_from` is asymmetric (which is surprising on its own).

tools/find_unsafe2/src/lib.rs
```diff
@@ -246,21 +289,96 @@ impl FunctionOutputs {
...
+            folded.insert(key.to_string(), value);
+        }
+    }
+    for v in folded.values_mut() {
+        v.calc_total_unsafe();
+    }
+    out.total_unsafe = folded.values()
+        .filter(|v| !v.is_ffi_entry_point)
+        .map(|v| v.total_unsafe)
+        .sum();
```
spernsteiner: Recalculating the totals doesn't belong in the folding function IMO

tools/find_unsafe2/src/lib.rs
```diff
@@ -327,6 +465,18 @@ fn ty_contains_raw_ptr(ty: Ty) -> usize {
                 Some(&RigidTy::Adt(adt, _)) => {
                     if matches!(&*adt.name(), "core::ptr::NonNull" | "std::ptr::NonNull") {
                         self.count += 1;
+                    } else if adt.krate().is_local && self.visited_adts.insert(adt) {
+                        // A pointer carried inside a local struct is still a pointer in the
+                        // signature.  Only local Adts are descended: std types reach `NonNull`
+                        // by field descent (`String` -> ... -> `NonNull`), and their pointer
+                        // fields are not the function's own unsafety surface.  Field types are
+                        // taken unsubstituted; concrete generic args are walked by
+                        // `super_visit` below, so substituting here would double-count.
```
spernsteiner: Why is this necessary?  We already prevent the agent from hiding raw pointers inside wrappers by stopping it from defining a new struct with a raw-pointer field.

tools/find_unsafe2/src/lib.rs
```diff
@@ -405,6 +555,25 @@ pub fn process(tcx: TyCtxt) -> Outputs {
         }
     };
 
+    let mut is_fn_ffi_entry_point = {
```
spernsteiner, line 506: We already have `is_ffi_entry_point` below; how is this new function different?

tools/find_unsafe2/src/lib.rs
```diff
@@ -419,46 +588,70 @@ pub fn process(tcx: TyCtxt) -> Outputs {
...
     let mut out = Outputs {
         total_unsafe: 0,
         fns: IndexMap::new(),
         types: IndexMap::new(),
+        unsafe_impls: IndexMap::new(),
     };
 
+    {
+        use rustc_hir::def::DefKind;
+        for item_id in tcx.hir_free_items() {
```
spernsteiner: We're trying to avoid using `TyCtxt` APIs as much as possible, on the theory that `rustc_public` / Stable MIR will be more stable across toolchain upgrades.  So this should use `all_trait_impls()` instead of accessing `tcx`.

tools/find_unsafe2/src/lib.rs
spernsteiner: Seems like every entry in the `unsafe_impls` map will always have the value 1.  If that's accurate, maybe it should be a set instead of a map.

tools/find_unsafe2/src/lib.rs
spernsteiner: This is converting from the `rustc_public` `DefId` type to the internal `DefId` type that can be used with `tcx`.  We have a bunch of little helpers that do this in order to access information that isn't exposed through `rustc_public`.

tools/find_unsafe2/src/bin/check_unsafe2.rs
spernsteiner: The `Some` case is using `Debug` (`{:?}`), so it will always be quoted.

tools/find_unsafe2/src/bin/check_unsafe2.rs
thedataking: Your suggestion is better since it allows new crates to be added as long as they're fully safe. 

OTOH, it doesn't look like `run_rewrite` currently carries forward new `Cargo.toml` files so either approach provides defense in depth.

tools/find_unsafe2/src/bin/check_unsafe2.rs
thedataking: Sounds like no action is required then. Resolving.

tools/find_unsafe2/src/lib.rs
thedataking: It let me resume a run where old steps didn't have this information. Removed.

tools/find_unsafe2/src/lib.rs
thedataking: Removed.

crisp/analysis.py
thedataking: If the agent is pressed to reduce unsafe ("you MUST make progress or this run will be terminated"), one way it can cheat is by jettisoning a crate from the workspace or using feature guards to compile out unsafe code so it is no longer compiled by `cargo check-unsafe2`. If the jettisoned code is not necessary for the tests to pass, I think the agent could get away with it.

crisp/analysis.py
thedataking: The agent runs `cargo check-unsafe2` to validate its changes and the agent can choose what working directory it does so from.  `FIND_UNSAFE2_SRC_DIR` makes sure the agent produces the same result as the subsequent `check_unsafe2`  step. Not sure this matters for the single-crate case; want me to drop it?

tools/find_unsafe2/src/lib.rs
thedataking: > Why is this necessary? We already prevent the agent from hiding raw pointers inside wrappers by stopping it from defining a new struct with a raw-pointer field.

The wrapper doesn't have to be new. It can reuse an existing struct emitted by `c2rust transpile`.

thedataking: Found an interesting case from previous `zlib` run:  https://github.com/immunant/zlib/blob/perl/crisp3/translated_rust/src/deflate.rs#L276

```rust
    /// Attach this owned allocation to an already-initialized deflate state.
    ///
    /// This is deliberately the one place where the eventual ownership
    /// facade recreates the legacy buffer handles.  It first proves that the
    /// state geometry and all four `Vec`s agree, so callers cannot derive a
    /// buffer view from a capacity advertised by an unrelated state.  The
    /// vectors are retained by `self` and are never resized after this handoff,
    /// keeping the established handles stable for the resumable legacy
    /// strategy engine.
    fn bind_state_buffers(&mut self, state: &mut crate::src::deflate::deflate_state) -> bool {
        if !self.matches_state(state) {
            return false;
        }
        state.window = ::std::sync::Arc::new(::core::sync::atomic::AtomicPtr::new(self.window.as_mut_ptr()));
        state.prev = ::std::sync::Arc::new(::core::sync::atomic::AtomicPtr::new(self.prev.as_mut_ptr()));
        state.head = ::std::sync::Arc::new(::core::sync::atomic::AtomicPtr::new(self.head.as_mut_ptr()));
        state.pending_buf = ::std::sync::Arc::new(::core::sync::atomic::AtomicPtr::new(self.pending_buf.as_mut_ptr()));
        true
    }
```

`bind_state_buffers` was introduced by the agent before this change, it is not in the C sources. 

`deflate_state` is an alias for `internal_state` (deflate.rs:79), which still carries four counted pointer fields: `window`, `prev`, `head`, `pending_buf`, each an `Arc<AtomicPtr<Bytef/Posf>>` (deflate.rs:104, 131, 133, 137).

I think this rule, as proposed, is too strict. I'm experimenting with a relaxed version where this check is only applied to safe functions so that yet-to-cleaned-up functions can carry raw pointers  so legitimate refactors do not get blocked inatvertently.

spernsteiner:
Rebased and cleaned up everything except 3cd5cde8196e25450ed3b68e2051a1b0c3ad65c0, which I dropped because I'm not convinced of its usefulness and there's a comment above saying it's too strict anyway.  There's a CI failure I need to fix regarding local file paths getting included in one of the snapshot tests.

spernsteiner:
Rebased again and made a small adjustment to have unsafe impl errors print the containing module path rather than the containing file.  This avoids including local file paths in snapshots, which were causing the previous CI failures.

I'm planning to merge this once CI passes.

PR #171, opened 2026-07-29
loop: recover review findings from op_history across restarts

thedataking:
Stacked on #169.

Rejected-review findings are made durable: the sticky reminder list is rebuilt from FAIL reviews in the op_history reflog at startup, so restarts stop repeating attempts a reviewer already rejected.


crisp/workflow.py
```diff
@@ -1390,6 +1431,26 @@ def do_ffi_review(
         if old_defs == new_defs:
             return True, None
 
+        # Cheap deterministic gate before the LLM review: catches signature
+        # and qualifier changes instantly, with precise feedback.
```
spernsteiner: Have we actually seen cases where the agent changes the signature of an FFI function?  This check seems a bit tricky, so we should omit it if possible.  If the only concern is qualifiers (e.g. removing `unsafe fn` from an FFI export), we're probably better off handling that in check-unsafe2 rather than doing this string matching.

thedataking: On the run before this change was added, we did 172 reviews, leading to 57 rejections. 9 of those reviews would have been short-circuited by this step. 

> If the only concern is qualifiers (e.g. removing unsafe fn from an FFI export), we're probably better off handling that in check-unsafe2 rather than doing this string matching.

8/9 checks dropped unsafe qualifier; we're saving ~5% of FFI review invocations for zlib.

Agreed that this sort of check would be easier to do from `check-unsafe2`.  Should we put a pin in this and triage the rest of this stack or do you think it is worth moving into `check_unsafe2`?

thedataking: I'm going to switch the FFI  reviewer to luna@xhigh (comparable intelligence to sol@medium). 
Should be cheaper & faster with the recent 80% price drop. It also decorrelates the models; the safety agent is terra, the reviewer being another model means they don't share the same blind spots. 

That' means we can drop this commit and not further complicate `check_unsafe2`. WDYT @spernsteiner?

PR #174, opened 2026-08-10
Add a script for graphing unsafe count over time

spernsteiner:
Useful for visualizing the rate of progress as unsafe count decreases throughout a CRISP run.  The script accepts multiple inputs to allow comparing several runs on a single graph.


scripts/graph_unsafe.py
```diff
@@ -0,0 +1,177 @@
...
+
+def parse_args():
+    ap = argparse.ArgumentParser(
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    ap.add_argument('--config', '-c', dest='config_path', default='crisp.toml')
+    ap.add_argument('--mvir-storage-dir')
+    ap.add_argument('--reflog-tag')
+    ap.add_argument('--print-interval', metavar='N', type=int, default=3600,
+        help='print unsafe counts at intervals of N seconds')
+    ap.add_argument('inputs', nargs='*', default='current',
```
ahomescu: Should the default be a list, i.e. `default=['current']`?

scripts/graph_unsafe.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def format_func(x, pos):
+        hours = int(x // 3600)
+        minutes = int((x % 3600) // 60)
+        seconds = int(x % 60)
+        return f'{hours}:{minutes:02}'
+
+
+    # Print summary stats
+    for x in itertools.count(step = args.print_interval):
```
ahomescu, line 172: This turns into an infinite loop if `args.print_interval` is non-positive. Add a check somewhere?

scripts/graph_unsafe.py
```diff
@@ -0,0 +1,177 @@
...
+    node = mvir.node(node_id)
+
+    history = get_history(mvir, node)
+    print(f'history: {len(history)} entries')
+
+    timestamp_map = {}
+    reflog_tag = args.reflog_tag or (node_id_arg if is_tag else 'current')
+    print(f'reading reflog of {reflog_tag!r}')
+    for re in mvir.tag_reflog(reflog_tag):
+        timestamp_map[re.node_id] = re.timestamp
```
ahomescu, line 74: Should this only set the value if it's unset? Otherwise it might overwrite an older timestamp with a newer one.

crisp/git.py
ahomescu: Is this dead code?

scripts/graph_unsafe.py
```diff
@@ -0,0 +1,177 @@
...
+        # Use elapsed time (in seconds) rather than an absolute `datetime` for
+        # the X axis, so we can compare results from different runs.
+        base_date = points[0][0]
+        ax.plot(
+            [(x[0] - base_date).total_seconds() for x in points],
+            [x[1] for x in points],
+            linestyle='-',
+        )
+    ax.grid(True)
+    ax.set_ylabel("Unsafe operations")
```
ahomescu: A legend and/or x axis label could be useful?

thedataking:
(Deleted: I ran into trouble because I accidentally checked out an earlier version of the script.)

scripts/graph_unsafe.py
thedataking: :+1: for legend, would be nice to add "zlib-baseline", "zlib-crisp5", etc. to graph and filename via command line argument.

spernsteiner: Added legend and X-axis label in 72525f42795b69ad124f7ec0e67f3a89efbe0378

crisp/git.py
spernsteiner: Removed dead code in 4f277ac5a65854bd8aca24bb87e60af50f0ac72b

scripts/graph_unsafe.py
spernsteiner: Added a comment describing the "take newest timestamp" behavior in 4f277ac5a65854bd8aca24bb87e60af50f0ac72b

scripts/graph_unsafe.py
spernsteiner: Added check for non-positive `print_interval` in 4f277ac5a65854bd8aca24bb87e60af50f0ac72b

scripts/graph_unsafe.py
spernsteiner: Fixed the default and squashed into 1219555074c976ef6c2a2523305666558405804c

scripts/graph_unsafe.py
```diff
@@ -0,0 +1,196 @@
+"""
+Graph unsafe count over time throughout the history of a refactoring run.
+
+Usage:
+    uv run --project /path/to/tractor-crisp --extra graph \
+        /path/to/tractor-crisp/scripts/graph_unsafe.py \
+        INPUTS...
+
```
thedataking, line 6: nit: It would be good to not have to pass `--project ...` because the `crisp` package lives in a sibling dir. 

What do you think about moving the script under `crisp` and registering it as a subcommand and making matplotlib non-optional so we can simply `uv run crisp graph-unsafe ...`? If we don't want a subcommand, we could also do something like:

```toml
[project.scripts]
crisp = "crisp.__main__:main"
crisp-graph-unsafe = "crisp.graph_unsafe:main"
```
if you're okay moving `graph_unsafe.py` under `crisp`.

thedataking (approved):
LGTM. 

Would be great to simplify how we run the script but not a blocker.

scripts/graph_unsafe.py
spernsteiner: Refactoring it to be a proper subcommand seemed a bit nontrivial, so instead I added a bit at the top to make it `uv run` itself if you call it directly with `python3 graph_unsafe.py`: 716b199906bec345b7159765cf38e5d801df714f

PR #175, opened 2026-08-11
crisp: enable a couple refactor passes to shorten code

fw-immunant:
Other passes might be also worthwhile, but these ones in particular seem like they should result in smaller code and plausibly less token usage when reading+rewriting this code via LLM subsequently.


fw-immunant:
Closing in favor of #176, since PRs from external clones of the repo can't access secrets needed for CI.

PR #176, opened 2026-08-11
crisp: enable a couple refactor passes to shorten code

fw-immunant:
Other passes might be also worthwhile, but these ones in particular seem like they should result in smaller code and plausibly less token usage when reading+rewriting this code via LLM subsequently.

### DIFF
--- a/crisp/__main__.py
+++ b/crisp/__main__.py
@@ -245,7 +245,7 @@ def do_main(args, cfg):
         n_code = w.transpile(
             n_c_code,
             src_loc_annotations=True,
-            refactor_transforms=("rename_unnamed", "reorganize_definitions"),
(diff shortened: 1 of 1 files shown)


spernsteiner (approved):
Seems simple enough

crisp/__main__.py
```diff
@@ -245,7 +245,7 @@ def do_main(args, cfg):
         n_code = w.transpile(
             n_c_code,
             src_loc_annotations=True,
-            refactor_transforms=("rename_unnamed", "reorganize_definitions"),
+            refactor_transforms=("rename_unnamed", "reorganize_definitions", "remove_redundant_casts", "remove_redundant_let_types"),
```
ahomescu, line 248: `remove_literal_suffixes` might also help a little after `remove_redundant_casts`, although we need it less than we used to.

ahomescu (approved):

